PR #2, opened 2024-04-30
Created an initial project structure

bretbrownjr:
This is a starting point based on
https://github.com/bretbrownjr/zerocode as of today. Actual decisions are still open to comment, but having someting to start from will be helpful for facilitating collaboration among the people sprinting to get an MVP for Beman this week.


bretbrownjr:
Note this doesn't have the same directory structure as #1, so we'll have to merge the approaches.

bretbrownjr:
Apologies for self-merging, but it looks like we need to do that to get github to enable actions fully.

PR #6, opened 2024-05-03
Improve ci

mguid65:
- Modify CI to use default images from github
    - Add windows, linux, and macos
    - Note: 
         - For Ninja, we have to fetch it from an external source
         - For Windows, we have to do some work to setup the msvc dev environment
         - For both of the above, these are accomplished by using public actions from the marketplace
             - This may not be appropriate and so we may have to do something different


.github/workflows/cxx.yml
```diff
@@ -1,83 +1,97 @@
-# Copyright © 2024 Bret Brown
-# SPDX-License-Identifier: MIT
-
-name: Test
+name: Test Multi-Platform
 
 on:
   push:
     branches: [ main ]
-    paths:
```
wusatosi, line 9: Please remove:
https://github.com/beman-project/exemplar/blob/d9145b52c9072a569d747547edfc7e03a1494809/.github/workflows/ci_tests.yml#L9-L10

If you intend to merge this change.

wusatosi:
Is there any update to this pr?
I wish to propose similar changes like this, is there a specific reason there's no update here?

@neatudarius @camio @bbrown105

mguid65:
I apologize, the reason for the lack of updates is that I forgot I made this PR.

We can either continue with proposed changes to this or we can discard it.

CMakePresets.json
```diff
@@ -0,0 +1,105 @@
...
+            "output": {
+                "outputOnFailure": true
+            },
+            "execution": {
+                "noTestsAction": "error",
+                "stopOnFailure": true
+            }
+        }
+    ]
+}
```
neatudarius, line 105:
```suggestion
}

```

neatudarius (commented):
@mguid65 , would you like to update your branch with latest main before continue with this review? Some changes are not applicable anymore.

e.g. Due to one usage examples, now the project requires C++20 or later.


Would you like to try to re-enable this example on your updated branch https://github.com/beman-project/exemplar/blob/main/examples/CMakeLists.txt#L4?

.github/workflows/cxx.yml
```diff
@@ -1,83 +1,97 @@
-# Copyright © 2024 Bret Brown
-# SPDX-License-Identifier: MIT
```
neatudarius, line 2: why do you want to remove these lines. I think we should keep them.

btw, @camio , should we add a mention for YAML files into https://github.com/beman-project/beman/blob/main/docs/beman-standard.md#file-contents?

wusatosi:
> @mguid65 , would you like to update your branch with latest main before continue with this review? Some changes are not applicable anymore.
> 
> e.g. Due to one usage examples, now the project requires C++20 or later.
> 
> 
> Would you like to try to re-enable this example on your updated branch https://github.com/beman-project/exemplar/blob/main/examples/CMakeLists.txt#L4?

I think this PR is abandoned, see: https://github.com/beman-project/exemplar/issues/27#issuecomment-2371527251, and the comment above.

I can fork this PR and pick this up.

neatudarius:
As per previous reply, I closed this PR to avoid more confusion, we can fork it later.

.github/workflows/cxx.yml
camio: > btw, @camio , should we add a mention for YAML files into https://github.com/beman-project/beman/blob/main/docs/beman-standard.md#file-contents?

Sounds like a promising idea. Would you be willing to create a PR or issue for it?

neatudarius: Yes. I will do it.

neatudarius: Merged https://github.com/beman-project/beman/pull/44

PR #10, opened 2024-05-12
:art: change from cxx/hxx to cpp/hpp

dustingooding:
Problem:
- The example uses cxx/hxx extensions.

Solution:
- Use the more popular cpp/hpp extensions.

---

Inspired by https://discourse.boost.org/t/beman-project-coding-guidelines/88/3


camio:
Great contribution. Thanks @dustingooding!

PR #13, opened 2024-06-28
Fix CMake module base filename

bretbrownjr:
This patch renames `example.cmake` to `example-config.cmake`. This fixes a bug in which `find_package(example)` fails to locate the CMake module shipped by this project.

See the upstream CMake documentation for `find_package` for details about which file basenames are accepted when searching for a CMake modules via `find_package`, but names like `example.cmake` are not on the search path and will not be discovered properly.

Using `example-config.cmake` resolves this issue.

### DIFF
--- a/src/example/CMakeLists.txt
+++ b/src/example/CMakeLists.txt
@@ -32,6 +32,7 @@ install(TARGETS example
     NAMELINK_COMPONENT libexample-dev
 )
 install(EXPORT example
+  FILE example-config.cmake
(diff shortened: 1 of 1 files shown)


bretbrownjr:
This sort of bug is difficult to test meaningfully inside the repo. It generally shows up when using this library from another project via packaging. I am working on Conan support in https://github.com/beman-project/beman/pull/21, which is how this bug was discovered.

camio (approved):

PR #26, opened 2024-09-23
Update CI dependency versions

wusatosi:
This pr updates the continuous integration tools' version.
This fixes the warning "The following actions uses node12 which is deprecated and will be forced to run on node16: actions/checkout@v2. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/" on the actions panel.
This pr also updates the continuous integration file so that it triggers when an update to its config is applied.


wusatosi:
Tested by manual trigger, see: https://github.com/wusatosi/exemplar/actions/runs/11001483129

camio (approved):
Awesome. Thanks for the PR!

Issue #14, opened 2024-07-18
Apply the Beman Standard to beman.exemplar


neatudarius:
Apply the [Beman Standard]( https://github.com/beman-project/beman/blob/main/docs/BEMAN_STANDARD.md) to Beman.Example and possible more more tweaks.

----------
# Updates related to the Beman Standard:
## License
- [x] LICENSE.APPROVED
- [x] LICENSE.APACHE_LLVM
- [x] LICENSE.CRITERIA
 
## General

- [x] LIBRARY.NAMES
- [x] REPOSITORY.NAME
- [x] REPOSITORY.CODEOWNERS

## Top-level

- [x] TOPLEVEL.CMAKE
- [x] TOPLEVEL.LICENSE
- [x] TOPLEVEL.README

## README.md

- [x] README.TITLE
- [x] README.PURPOSE
- [x] README.IMPLEMENTS

## CMake

- [x] CMAKE.DEFAULT
- [x] CMAKE.PROJECT_NAME
- [x] CMAKE.LIBRARY_NAME
- [x] CMAKE.LIBRARY_ALIAS
- [x] CMAKE.TARGET_NAMES
- [ ] CMAKE.CONFIG
- [x] CMAKE.SKIP_TESTS
- [x] CMAKE.AVOID_PASSTHROUGHS

## Directory Layout

- [x] DIRECTORY.INTERFACE_HEADERS
- [x] DIRECTORY.IMPLEMENTATION_HEADERS
- [x] DIRECTORY.SOURCES
- [ ]  DIRECTORY.TESTS
- [ ]  DIRECTORY.EXAMPLES
- [ ]  DIRECTORY.DOCS
- [ ]  DIRECTORY.PAPERS

## File contents

- [ ]  FILE.NAME
- [ ] FILE.TEST_NAMES
- [x] FILE.LICENSE_ID
- [x] FILE.COPYRIGHT

## C++

- [x] CPP.NAMESPACE

# Updates related to docs and usage example:
- [x] Add API example
- [x] Add basic unit test and framework integration with Gtest.
- [x] Add usage example in `examples/`

# Updates related to CI/infrastructure
- [x] Compiler support on CI
- [x] Add `.gitignore`.
- [x] Add `.github/CODEOWNERS`.
- [x] Add badge with CI tests status in `README.md`.

neatudarius:
Most of the issues are implemented in #15

bretbrownjr:
[...]

PR #31, opened 2024-09-25
More tweaks after standard updates

neatudarius:
Missing changed from #14 

Updates:
- [ ]  Apply CMAKE.CONFIG

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,18 +4,22 @@
 cmake_minimum_required(VERSION 3.23)
 
 project(
-  beman.exemplar # CMake Project Name, which is also the name of the top-level
-                 # targets (e.g., library, executable, etc.).
+  beman.exemplar
   DESCRIPTION "A Beman library exemplar"
+  VERSION 0.0.0
   LANGUAGES CXX)
 
+include(CMakePackageConfigHelpers)
 include(CTest)
 include(FetchContent)
+include(GNUInstallDirs)
+
+set(TARGETS_EXPORT_NAME ${CMAKE_PROJECT_NAME}Targets)
+set(INSTALL_CONFIGDIR ${CMAKE_INSTALL_LIBDIR}/cmake)
 
 if(BUILD_TESTING)
   enable_testing()
 
-  # Fetch GoogleTest
   FetchContent_Declare(
     googletest
     GIT_REPOSITORY https://github.com/google/googletest.git
@@ -25,5 +29,19 @@ if(BUILD_TESTING)
 endif()
 
 add_subdirectory(src/beman/exemplar)
-
 add_subdirectory(examples)
+
+write_basic_package_version_file(
+  ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_PROJECT_NAME}ConfigVersion.cmake
+  VERSION ${PROJECT_VERSION}
+  COMPATIBILITY AnyNewerVersion)
+
+configure_package_config_file(
+  "cmake/Config.cmake.in"
+  ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_PROJECT_NAME}Config.cmake
+  INSTALL_DESTINATION ${INSTALL_CONFIGDIR})
+
+install(
+  FILES ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_PROJECT_NAME}Config.cmake
+        ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_PROJECT_NAME}ConfigVersion.cmake
+  DESTINATION ${INSTALL_CONFIGDIR})
--- a/README.md
+++ b/README.md
@@ -152,7 +152,8 @@ $ cmake -B build -S . -DCMAKE_CXX_STANDARD=20
 -- Check for working CXX compiler: /usr/bin/c++ - skipped
 -- Detecting CXX compile features
 -- Detecting CXX compile features - done
--- Configuring done (0.1s)
+...
+-- Configuring done (2.9s)
 -- Generating done (0.0s)
 -- Build files have been written to: /path/to/repo/build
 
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -25,5 +29,19 @@ if(BUILD_TESTING)
 endif()
 
 add_subdirectory(src/beman/exemplar)
-
 add_subdirectory(examples)
+
+write_basic_package_version_file(
```
neatudarius, line 34: Inspect current behavior from https://github.com/beman-project/execution26/blob/main/CMakeLists.txt and https://github.com/beman-project/optional26/blob/main/CMakeLists.txt. 

I tried to extract only the minimum required code to do the install.

Expected tree:

```bash
@pc:~/git/Beman/exemplar$ cmake --install build --prefix /opt/beman.exemplar
-- Install configuration: ""
-- Installing: /opt/beman.exemplar/lib/libbeman.exemplar.a
-- Installing: /opt/beman.exemplar/include
-- Installing: /opt/beman.exemplar/include/beman
-- Installing: /opt/beman.exemplar/include/beman/exemplar
-- Installing: /opt/beman.exemplar/include/beman/exemplar/identity.hpp
-- Installing: /opt/beman.exemplar/lib/cmake/beman.exemplarConfig.cmake
-- Installing: /opt/beman.exemplar/lib/cmake/beman.exemplarConfigVersion.cmake
dariusn@pc:~/git/Beman/exemplar$ tree /opt/beman.exemplar
/opt/beman.exemplar
├── include
│   └── beman
│       └── exemplar
│           └── identity.hpp
└── lib
    ├── cmake
    │   ├── beman.exemplarConfig.cmake
    │   └── beman.exemplarConfigVersion.cmake
    └── libbeman.exemplar.a

6 directories, 4 files
```

I would like your feedback and decide a standard pattern, which will be applied initially to beman.examplar, and eventually mirror it into other 2 repos.

camio: > [W]e should NOT export...docs

I'm okay with this for now. Eventually, we should export docs at installation time to the `docs/beman.short_name/` directory, but we need to get our docs story figured out first.

> @camio proposed to add a subdirectory for cmake configs

Yes, this is the convention for the ArchLinux distribution and probably others as well. See [openjpeg2's contents](https://archlinux.org/packages/extra/x86_64/openjpeg2/) for an example. [Config Mode's Search Procedure](https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure) includes a table indicating how *Config.cmake files are found.

CMakeLists.txt
```diff
@@ -4,18 +4,22 @@
...
   DESCRIPTION "A Beman library exemplar"
+  VERSION 0.0.0
   LANGUAGES CXX)
 
+include(CMakePackageConfigHelpers)
 include(CTest)
 include(FetchContent)
+include(GNUInstallDirs)
+
+set(TARGETS_EXPORT_NAME ${CMAKE_PROJECT_NAME}Targets)
```
bretbrownjr, line 17: I'd rather just inline this below instead of making a variable that's used only once.

CMakeLists.txt
```diff
@@ -25,5 +29,19 @@ if(BUILD_TESTING)
 endif()
 
 add_subdirectory(src/beman/exemplar)
-
 add_subdirectory(examples)
+
+write_basic_package_version_file(
+  ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_PROJECT_NAME}ConfigVersion.cmake
```
bretbrownjr, line 35:
```suggestion
  ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_PROJECT_NAME}-config-version.cmake
```
Just because `lower.caseThenCamelCase.cmake` is not how anyone would choose to name a file. See [upstream docs](https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html) on support for kebab-case filenames.

CMakeLists.txt
```diff
@@ -25,5 +29,19 @@ if(BUILD_TESTING)
...
+  VERSION ${PROJECT_VERSION}
+  COMPATIBILITY AnyNewerVersion)
+
+configure_package_config_file(
+  "cmake/Config.cmake.in"
+  ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_PROJECT_NAME}Config.cmake
```
bretbrownjr, line 41:
```suggestion
  ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_PROJECT_NAME}-config.cmake
```

See the comment about `*ConfigVersion.cmake`

CMakeLists.txt
```diff
@@ -25,5 +29,19 @@ if(BUILD_TESTING)
 endif()
 
 add_subdirectory(src/beman/exemplar)
-
 add_subdirectory(examples)
+
```
bretbrownjr, line 33: Shouldn't all the library metadata (i.e., `exemplar-config.cmake`, etc.) go in the `src/beman/exemplar` subdir?

CMakeLists.txt
```diff
@@ -25,5 +29,19 @@ if(BUILD_TESTING)
...
+write_basic_package_version_file(
+  ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_PROJECT_NAME}ConfigVersion.cmake
+  VERSION ${PROJECT_VERSION}
+  COMPATIBILITY AnyNewerVersion)
```
bretbrownjr, line 37:
```suggestion
  COMPATIBILITY ExactVersion)
```

Unless we're doing something specific to guarantee some ABI compatibility across versions of a Beman project, we should be conservative about version matching. On the contrary, I fully expect API and ABI breaks across releases of Beman projects since they're by definition experimental and prone to design changes.

We could *maybe* revisit this choice on a case-by-case basis for projects that a "stable" status. But unless they're spelling their implementations with ABI tags and such, I wouldn't advise people bother.

CMakeLists.txt
```diff
@@ -25,5 +29,19 @@ if(BUILD_TESTING)
...
+
+configure_package_config_file(
+  "cmake/Config.cmake.in"
+  ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_PROJECT_NAME}Config.cmake
+  INSTALL_DESTINATION ${INSTALL_CONFIGDIR})
+
+install(
+  FILES ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_PROJECT_NAME}Config.cmake
```
bretbrownjr, line 45:
```suggestion
  FILES ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_PROJECT_NAME}-config.cmake
```

CMakeLists.txt
```diff
@@ -25,5 +29,19 @@ if(BUILD_TESTING)
...
+        ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_PROJECT_NAME}ConfigVersion.cmake
```
bretbrownjr, line 46:
```suggestion
        ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_PROJECT_NAME}-config-version.cmake
```

README.md
```diff
@@ -221,9 +224,12 @@ $ tree /opt/beman.exemplar
 │       └── exemplar
 │           └── identity.hpp
 └── lib
+    ├── cmake
+    │   ├── beman.exemplarConfig.cmake
+    │   └── beman.exemplarConfigVersion.cmake
```
bretbrownjr, line 229:
```suggestion
    │   ├── beman.exemplar-config.cmake
    │   └── beman.exemplar-config-version.cmake
```

Also, what happened to the file referenced by this bit in the config file?
`include("${CMAKE_CURRENT_LIST_DIR}/@TARGETS_EXPORT_NAME@.cmake")`

bretbrownjr (changes requested):
I have a few suggestions here and there. The most important changes I'd like to request are:

* I suspect we're missing the `Targets.cmake` file from the installation logic, given the `tree` command doesn't note it
* I suspect we're missing some sort of test to confirm the packaging works. I guess that could be another PR, but it would be nice to at least manually test that installing into a staging dir, manipulating `CMAKE_PREFIX_PATH`, and then calling `find_package(exemplar REQUIRED)` in an example works well. For what it's worth, Conan ships with a test that does this.
* We'll want to do exact version matching or just ignore versions altogether (both work for me). Trying to support semantic versioning for locating a built binary is not a recipe for success. That's only really appropriate for intentionally ABI stable projects like `libc++` or various C libraries. Certainly not Beman libraries, which are disclaiming stability, at least until they reach the "stable" status.

Minor general note: At some point in recent refactoring, we also lost the support for `install` `COMPONENTS`. We should add that back in, but it can be another PR.

FYI @steve-downey. I'm guessing we aren't using install components in optional26 either.

wusatosi:
@neatudarius Mind if I contribute directly to this PR?

neatudarius:
> @neatudarius Mind if I contribute directly to this PR?

Sure, please contribute

wusatosi:
> > @neatudarius Mind if I contribute directly to this PR?
> 
> Sure, please contribute

I am in my finals week, I no longer have extra capacity to work on this.

camio:
There's was a review of this PR on October 1st, but no updates since. Should we close this PR and re-open it when someone wants to drive it forward?

neatudarius:
> There's was a review of this PR on October 1st, but no updates since. Should we close this PR and re-open it when someone wants to drive it forward?

I will finish it this week. Sorry!

neatudarius:
N/A, dropping this PR.

PR #35, opened 2024-09-28
Remove `return 0` from main

### LINKED ISSUE #33: Examples should not `return 0;` from main
From [cppreference](https://en.cppreference.com/w/cpp/language/main_function):

> The body of the main function does not need to contain the [return statement](https://en.cppreference.com/w/cpp/language/return): if control reaches the end of main without encountering a return statement, the effect is that of executing return 0;.

This change improves readability by omitting needless code.

wusatosi:
As requested by #33.
closes #33

Edit: I don't really like this as well, but just creating one so if everyone agrees a pr is ready, feel free to close this lol.

### DIFF
--- a/examples/identity_as_default_projection.cpp
+++ b/examples/identity_as_default_projection.cpp
@@ -84,6 +84,4 @@ int main()
     print_std("\tpairs with   std: ", pairs,
               [](const auto &p)
               { return std::to_string(p.n) + ':' + p.s; });
-
-    return 0;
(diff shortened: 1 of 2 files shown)


neatudarius (commented):
Let's wait the discussion from #33 to be finished.

Otherwise, LGTM.

wusatosi:
Look like this is not going to be implemented, closing for now

PR #37, opened 2024-09-30
Disable installing gtest when `cmake --install`

### LINKED ISSUE #32: gtest and gmock are inadvertantly installed using older versions of CMake
[`FechContent`](https://cmake.org/cmake/help/latest/module/FetchContent.html)'s `EXCLUDE_FROM_ALL` keyword in CMake 3.30+ has the impact of ignoring `install` commands in the dependency project. However, this behavior does not manifest in earlier CMake versions which results in exemplar's gtest dependency being installed when CMake's install target is executed.

CMake 3.30 came out recently ([July 2024](https://github.com/Kitware/CMake/releases/tag/v3.30.0)), so forcing users to upgrade to use beman.exemplar is a bit much to ask.

My suggested fix is to set [`INSTALL_GTEST`](https://github.com/google/googletest/blob/6dae7eb4a5c3a169f3e298392bff4680224aa94a/CMakeLists.txt#L20) to `FALSE` using a [`block`](https://cmake.org/cmake/help/latest/command/block.html) that contains the `FetchContent` call.

wusatosi:
closes #32 .

This pr updates the CMake file, and disables installing `gtest` headers when running `cmake --install`.

Note that this will require cmake 3.25 due to use of [`block`](https://cmake.org/cmake/help/latest/command/block.html).

See: https://github.com/beman-project/exemplar/pull/37/checks#step:6:320 for effect of this pr.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 # TODO Darius: Find CMake minimum required version.
-cmake_minimum_required(VERSION 3.23)
+cmake_minimum_required(VERSION 3.25)
 
 project(
   beman.exemplar # CMake Project Name, which is also the name of the top-level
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 # TODO Darius: Find CMake minimum required version.
```
wusatosi, line 3: @neatudarius is this line still relevant?

wusatosi:
The use of block is nice but isn't necessary, I don't think there's any chance we have conflict on the name `INSTALL_GTEST`

wusatosi:
Explicitly requesting review from @camio.

CMakeLists.txt
neatudarius: If this is the new minimum required version,  we can remove the todo

neatudarius (approved):
LGTM from my side

neatudarius (approved):

camio (approved):
Looks perfect!

PR #39, opened 2024-10-01
Install debug build binaries to a separate folder

ComicSansMS:
This will mirror the directory structure used by vcpkg:
* Debug binaries will get installed to a subdirectory debug/lib for ARCHIVE and LIBRARY artifacts and debug/bin for RUNTIME artifacts
* All other configurations install binaries directly to lib and bin respectively
* Other artifacts go to the same include/share/... directories for all configurations, including Debug


camio:
Thanks for the contribution @ComicSansMS! Could you add to the `.github/workflows/ci_tests.yml` a check so we can verify this works properly (at least by eye)? I'm thinking a Ninja MultiConfig including both debug and release builds.

ComicSansMS:
> Thanks for the contribution @ComicSansMS! Could you add to the `.github/workflows/ci_tests.yml` a check so we can verify this works properly (at least by eye)? I'm thinking a Ninja MultiConfig including both debug and release builds.

So, the easiest way to do this is to just have all builds always build both Debug and Release configurations. I added that change to the PR.

To properly ensure that this works, we will eventually need a CI job that mimics a client application consuming the exemplar library in different configurations. I will leave that for another PR though, as it requires a completely new CI job that will be quite a bit more complex than what we have now.

camio (approved):
Looks great! Thanks again for the contribution.

PR #42, opened 2024-10-02
Simplify CMakeLists.txt

camio:
The `block()` isn't needed since we can set CMake options
via. CMAKE_ARGS.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,18 +14,17 @@ include(FetchContent)
 if(BUILD_TESTING)
   enable_testing()
 
-  block()
-    # Disable installing google test dependency on cmake --install
-    set(INSTALL_GTEST OFF)
-
-    # Fetch GoogleTest
-    FetchContent_Declare(
-      googletest
-      GIT_REPOSITORY https://github.com/google/googletest.git
-      GIT_TAG f8d7d77c06936315286eb55f8de22cd23c188571 # release-1.14.0
-      EXCLUDE_FROM_ALL CMAKE_ARGS -DBUILD_TESTING=OFF)
-    FetchContent_MakeAvailable(googletest)
-  endblock()
+  # Fetch GoogleTest
+  FetchContent_Declare(
+    googletest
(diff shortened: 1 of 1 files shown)


neatudarius (approved):

camio:
This, surprisingly, didn't work as gtest files are being installed again with this change.

DeveloperPaul123:
@camio Are you using the necessary version of CMake (`3.30+`)? It could be that we also need to change `cmake_minimum_required(VERSION 3.25)` to `cmake_minimum_required(VERSION 3.25..3.30.4)` to properly set all the policies.

wusatosi:
I don't see this being supported? 

https://cmake.org/cmake/help/latest/module/FetchContent.html

DeveloperPaul123:
Nice catch @wusatosi !

camio:
Okay, I think I figured this out. Thanks @wusatosi and @DeveloperPaul123 for your help.

1. `CMAKE_ARGS` is ignored in `FetchContent_Declare`. The documentation states "The configure, build, install, and test steps are explicitly disabled, so options related to those steps will be ignored." and `CMAKE_ARGS` list listed as a "configure" option.
2. The `add_subdirectory` call is made in `FetchContent_MakeAvailable` so it makes sense to have the `FetchContent_Declare` call outside of the block.
3. The use of `CMAKE_ARGS` to disable `BUILD_TESTING` appeared to work only because GoogleTest disables tests by default and uses a different variable to control testing.
4. Although `BUILD_TESTING` doesn't do anything in this case, I think we should keep it there as other dependency projects will use the standard variable for this.

@wusatosi , @DeveloperPaul123 can you give this PR another look now? I've verified in the CI output that this works as expected.

neatudarius (approved):

wusatosi (approved):
LGTM! 

Turns out we still need the `block` directive at the end :( 
But its nice to know the BUILD_TESTING option is off correctly for google test now!.

DeveloperPaul123 (approved):
Looks good to me! Nice work 👍

PR #43, opened 2024-10-03
Implement schedule fault notification

### LINKED ISSUE #36: There should be some notification mechanism when scheduled build-test ci fails
Scheduled build-and-test introduced in #28 is triggering correctly.
But failure to routine build-and-test can be silent, e. g. https://github.com/beman-project/exemplar/actions/runs/11109516669 .
(technically GitHub pushes a notification to the latest person that edited cron syntax, but this is clearly insufficient).

There needs to be some notification mechanism to notify the project owner/ contributors that the latest build-and-test fails.

Opening a GitHub issue should be sufficient enough?

wusatosi:
closes #36 . 

Scheduled build-and-test introduced in #28 is triggering correctly.
But failure at routine build-and-test can be silent, e. g. https://github.com/beman-project/exemplar/actions/runs/11109516669 .

This pr creates a new issue when scheduled build-and-test fails.
To avoid spamming the issue tracker, it will append to an existing open issue with a comment if an previously opened issue exist and is active.

Tested on my own fork, see: https://github.com/wusatosi/exemplar/issues/6

### DIFF
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -69,3 +69,32 @@ jobs:
         run: |
           docker run ${{ matrix.cfg.id }} ctest --test-dir build -C Release
           docker run ${{ matrix.cfg.id }} ctest --test-dir build -C Debug
+
+  create-issue-when-fault:
+    runs-on: ubuntu-latest
+    needs: [test]
+    if: failure() && github.event_name == 'schedule'
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


steve-downey (approved):
Looks good to me. 
Does the leading whitespace get trimmed properly in the resulting comments?

wusatosi:
> Looks good to me. Does the leading whitespace get trimmed properly in the resulting comments?

Yes, see: https://github.com/wusatosi/exemplar/issues/6

PR #7, opened 2024-05-03
Add vcpkg configuration files

hallfox:
This should be the minimal set of files to support a vcpkg-backed toolchain. Further instructions on how one can add more dependencies for to be fetched by vcpkg can be found at the [getting started page on vcpkg's docs](https://learn.microsoft.com/en-us/vcpkg/get_started/get-started?pivots=shell-cmd#3---add-dependencies-and-project-files).

PR #44, opened 2024-10-04
Create CMake preset

### LINKED ISSUE #17: Add CMake preset support

wusatosi:
closes #17 .

This is a work in process seeking feedback, currently only included gcc related workflows.
Needs feedback on which compiler and what flags I should include/ exclude.
Let's keep the list of presets short (https://github.com/beman-project/exemplar/issues/17#issuecomment-2362899915) so we don't strain the CI system too much.

I omitted `-Werror` in #5 as this flag isn't indicated elsewhere (e.g. README).

There's two workflow presets:
- `gcc-debug`, (`-g` with all non-conflicting sanitizer included, see: https://github.com/beman-project/exemplar/issues/17#issuecomment-2377042239)
- `gcc-release`, (`-g` with `-O3`)

Assumes C++ 20 per: https://github.com/beman-project/exemplar/pull/38#issuecomment-2391421004 , but will monitor the progress happening at #38 . Please forward minimum version discussion to that thread unless it should be specific to a workflow.

CI test is only meant to test if the workflow config is functional, main testing should be done in the main build & test matrix, thus there's no cross-platform test. We could merge the two but that will need future discussion.

This is also a good peak into docker-less build & test for CI.

Please test this locally, I have experienced the test process hanging, possibly due to the excessive amount of sanitizer enabled. We may need to include a warning in the README to advise turning off/ splitting specific performance-impacting sanitizer on bigger projects.

Related pr: #7

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,0 +1,121 @@
+{
+  "version": 6,
+  "configurePresets": [
+    {
+      "name": "_root-config",
+      "hidden": true,
+      "generator": "Ninja",
+      "binaryDir": "${sourceDir}/build/${presetName}",
+      "cacheVariables": {
+        "CMAKE_CXX_STANDARD": "20"
+      }
+    },
+    {
+      "name": "_debug-base",
+      "hidden": true,
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Debug",
+        "CMAKE_CXX_FLAGS": "-g -fsanitize=address -fsanitize=pointer-compare -fsanitize=pointer-subtract -fsanitize=leak -fsanitize=undefined"
+      }
+    },
+    {
+      "name": "_release-base",
+      "hidden": true,
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "RelWithDebInfo",
+        "CMAKE_CXX_FLAGS": "-g -O3"
+      }
+    },
+    {
+      "name": "gcc-debug",
+      "displayName": "GCC Debug Build",
+      "inherits": [
+        "_root-config",
+        "_debug-base"
+      ],
+      "cacheVariables": {
+        "CMAKE_CXX_COMPILER": "g++"
+      }
+    },
+    {
+      "name": "gcc-release",
+      "displayName": "GCC Release Build",
+      "inherits": [
+        "_root-config",
+        "_release-base"
+      ],
+      "cacheVariables": {
+        "CMAKE_CXX_COMPILER": "g++"
+      }
+    }
+  ],
+  "buildPresets": [
+    {
+      "name": "gcc-debug",
+      "configurePreset": "gcc-debug"
+    },
+    {
+      "name": "gcc-release",
+      "configurePreset": "gcc-release"
+    }
+  ],
+  "testPresets": [
+    {
+      "name": "_test_base",
+      "hidden": true,
+      "output": {
+        "outputOnFailure": true
+      },
+      "execution": {
+        "noTestsAction": "error",
+        "stopOnFailure": true
+      }
+    },
+    {
+      "name": "gcc-debug",
+      "inherits": "_test_base",
+      "configurePreset": "gcc-debug"
+    },
+    {
+      "name": "gcc-release",
+      "inherits": "_test_base",
+      "configurePreset": "gcc-release"
+    }
+  ],
+  "workflowPresets": [
+    {
+      "name": "gcc-debug",
+      "steps": [
+        {
+          "type": "configure",
+          "name": "gcc-debug"
+        },
+        {
+          "type": "build",
+          "name": "gcc-debug"
+        },
+        {
+          "type": "test",
+          "name": "gcc-debug"
+        }
+      ]
+    },
+    {
+      "name": "gcc-release",
+      "steps": [
+        {
+          "type": "configure",
+          "name": "gcc-release"
+        },
+        {
+          "type": "build",
+          "name": "gcc-release"
+        },
+        {
+          "type": "test",
+          "name": "gcc-release"
+        }
+      ]
+    }
+  ]
+}
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -15,7 +15,7 @@
       "hidden": true,
       "cacheVariables": {
         "CMAKE_BUILD_TYPE": "Debug",
-        "CMAKE_CXX_FLAGS": "-g -fsanitize=address -fsanitize=pointer-compare -fsanitize=pointer-subtract -fsanitize=leak -fsanitize=undefined"
+        "CMAKE_CXX_FLAGS": "-fsanitize=address -fsanitize=pointer-compare -fsanitize=pointer-subtract -fsanitize=leak -fsanitize=undefined"
       }
     },
     {
@@ -23,7 +23,7 @@
       "hidden": true,
       "cacheVariables": {
         "CMAKE_BUILD_TYPE": "RelWithDebInfo",
-        "CMAKE_CXX_FLAGS": "-g -O3"
+        "CMAKE_CXX_FLAGS": "-O3"
       }
     },
     {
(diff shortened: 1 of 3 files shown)


bretbrownjr:
It could be a separate PR, but I expect especially for introducing a new, friendly workflow, we'll want to make sure we have the workflow documented for potential contributors and users.

CMakePresets.json
```diff
@@ -0,0 +1,121 @@
...
+      "cacheVariables": {
+        "CMAKE_CXX_STANDARD": "20"
+      }
+    },
+    {
+      "name": "_debug-base",
+      "hidden": true,
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Debug",
+        "CMAKE_CXX_FLAGS": "-g -fsanitize=address -fsanitize=pointer-compare -fsanitize=pointer-subtract -fsanitize=leak -fsanitize=undefined"
```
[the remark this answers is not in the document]
wusatosi: Nice catch, fixed!

CMakePresets.json
```diff
@@ -0,0 +1,121 @@
...
+      }
+    },
+    {
+      "name": "_release-base",
+      "hidden": true,
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "RelWithDebInfo",
+        "CMAKE_CXX_FLAGS": "-g -O3"
```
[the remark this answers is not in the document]
wusatosi: Nice catch, fixed!

camio (approved):
I tried this on an ubuntu machine and it's a good improvement.

In testing this I used `cmake --list-presets`, `cmake --workflow --preset gcc-release`, and `cmake --workflow --preset gcc-debug`. All worked as expected on an Ubuntu VM.

I tried this on MacOS and got an error. The reason is that MacOS has a `gcc` executable in the path that is actually clang under the hood...one that doesn't support the `-fsanitize=leak` option. Clang presets will help mitigate this issue, but it's good to be aware of in case we want to attempt better error messages in the future.

wusatosi:
> I tried this on an ubuntu machine and it's a good improvement.
> 
> In testing this I used `cmake --list-presets`, `cmake --workflow --preset gcc-release`, and `cmake --workflow --preset gcc-debug`. All worked as expected on an Ubuntu VM.
> 
> I tried this on MacOS and got an error. The reason is that MacOS has a `gcc` executable in the path that is actually clang under the hood...one that doesn't support the `-fsanitize=leak` option. Clang presets will help mitigate this issue, but it's good to be aware of in case we want to attempt better error messages in the future.

Thank you for looking over this PR. Should I include this in the documentation?

camio:
>> I tried this on MacOS and got an error.
> Should I include this in the documentation?

I don't think that's necessary. Let's see if others run into the same issue first. It feels kinda niche to me, especially if we include a clang preset.

wusatosi:
> It could be a separate PR, but I expect that especially for introducing a new, friendly workflow, we'll want to make sure we have the workflow documented for potential contributors and users.

What would that workflow be different from `gcc-debug`?

wusatosi:
> [It's not necessary to document failure to run gcc-debug on macos due to symlinking gcc with clang]

Noted! I've updated the documentation.
This pr is ready to merge if only this two presets is desired.

I could add clang-debug/release to this pr too, or I could also do this via another PR.

CMakePresets.json
```diff
@@ -0,0 +1,121 @@
...
+      },
+      "execution": {
+        "noTestsAction": "error",
+        "stopOnFailure": true
+      }
+    },
+    {
+      "name": "gcc-debug",
+      "inherits": "_test_base",
+      "configurePreset": "gcc-debug"
```
steve-downey, line 77: Not for this PR, but it just occurred to me that we might leverage the ninja-multiconfig generator to keep the list of configurePresets small? That is have one that configures Debug, Release, and possibly various sanitizer flavors? This might help keep the combinatorics down.

wusatosi: I don't think we are enumerating sanitizer flags here in the cmake preset file. As far as I've seen in prior conversations this preset is only meant to be a short hand for common configuration.

CMakePresets.json
```diff
@@ -0,0 +1,121 @@
+{
+  "version": 6,
+  "configurePresets": [
+    {
+      "name": "_root-config",
+      "hidden": true,
+      "generator": "Ninja",
```
wusatosi, line 7: Should this be `Ninja Multi-Config` as updated to [Dockerfile](https://github.com/beman-project/exemplar/pull/39/files#diff-74d629d3c2ebfd97a4acab7c9c2cdfcae721b98e92873131bac63b6454cdad46L29-R29) in #39 ?

@camio @ComicSansMS

camio: Yes, exactly. Feel free to do that in a separate PR though because this file will have to be refactored since there will be only one `gcc` preset needed after that.

wusatosi: I am not familiar with that, I will need to look into it. let's ship this PR then!

README.md
```diff
@@ -134,9 +134,7 @@ apt-get install                           \
 This project strives to be as normal and simple a CMake project as possible. This build workflow in particular will work, producing a static `libbeman.exemplar.a` library, ready to package with its headers:
 
 ```shell
-cmake -B build -S . -DCMAKE_CXX_STANDARD=20
-cmake --build build
-ctest --test-dir build
+cmake --workflow --preset gcc-debug
 cmake --install build --prefix /opt/beman.exemplar
```
camio, line 138: I suggest also including `gcc-release` since that's most likely what folks will want to install. I also had to change the install command to point to the right build directory.

```suggestion
cmake --workflow --preset gcc-release
cmake --install build/gcc-release --prefix /opt/beman.exemplar
```

README.md
```diff
@@ -134,9 +134,7 @@ apt-get install                           \
...
 ```
```
camio, line 140: Note that the "verbose logs" sections below need to be updated as well.

On the other hand, I'd be happier to see them removed entirely. I've used beman.exemplar for a few different projects now and it's a pain to have to update those sections.

The "Disable tests build" section also needs to be updated.

wusatosi: Thank you for double checking! Sorry I missed those sections. I will have them updated later today. !!

wusatosi: Should I simply remove the verbose logs then?

wusatosi: Updated cmake command and output in 84305172eb1af8ef7b7b7d0b9b17f0f3a5aa7c01 .

wusatosi:
In the README, we includes [support for turning off build test](https://github.com/beman-project/exemplar?tab=readme-ov-file#how-to-build-bemanexemplar). However, [`cmake` does not support passing additional parameters](https://gitlab.kitware.com/cmake/cmake/-/issues/24073)  [till 3.25](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/7825).

How should I update the documentations?
Should I
1. Remove the section entirely
2. Add a no build test variant to the preset
3. Indicate that workflow will not support this, and include commands to manually build the project with `BUILD_TEST` flag?

@camio , cc @bretbrownjr as that paragraph was introduced in #15 .

JeffGarland:
This looks like it's ready to close?

wusatosi:
> This looks like it's ready to close?

There's a pending documentation change, see above [comment](https://github.com/beman-project/exemplar/pull/44#issuecomment-2424171988).

JeffGarland:
There's a lot of PR's open here so hopefully we can close them soon :)

wusatosi:
> There's a lot of PR's open here so hopefully we can close them soon :)

#48 should be ready, can you review and merge it?

camio (approved):

camio:
I think this is ready to merge.

> How should I update the documentations?
> Should I
> 
> * Remove the section entirely

I think this is a good idea. Can be done in a different PR.

> * Add a no build test variant to the preset

I don't think we need a preset for this as it is a specialized need.

> Indicate that workflow will not support this, and include commands to manually build the project with BUILD_TEST flag?

I think this only needs to be discussed in a "how to incorporate in another project" section which needs more work anyway (e.g. a FetchContent example)

PR #34, opened 2024-09-27
Introduce basic linting infrastructure

wusatosi:
Introduce pre-commit based linting infrastructure:

- pre-commit **formats** extra white space, extra line at end-of-file, C++ files, and CMake files. It **flags** linting error for Markdown.
- CI does not push formatted change to pr/ main branch.
- Pre-commit is portable (even with `clang-format`) and does not pollute global scope, thus no virtual environment facing files (like `requirements.txt`) is included. This is tested with an empty GitHub Codespace.

Action Items:
- [x] setup a basic pre-commit infrastructure
- [x] setup a basic clang-format file, prefer discussion at #29 .
- [x] setup clang-format based formatting, see: 92b62736171fa9fd7692fc555f483ec85f0cdb5e for format diff
~~setup clang-tidy linting~~: may be out of scope, see discussion below: https://github.com/beman-project/exemplar/pull/34#discussion_r1784958937
- [x] setup cmake formatting (use [`gersemi`](https://github.com/BlankSpruce/gersemi), see: f3557fc485517922a656653a38c0a1c3fa662b60 for format diff)
- [x] setup markdown linting (via [`markdown-cli`](https://github.com/igorshubovych/markdownlint-cli), see: 1ca633837d4ecca08c58fb84c098cff4ed193fef for format diff)
- [x] setup ci check for linting
- [x] update project so it passes the linting test
- [ ] Update documentation (README)

Reference implementation: https://github.com/beman-project/optional26/pull/52
- Note that this PR diverges with reference implementation greatly:
- There's less "hooks" (linting rules)
- Uses different & updated tool chains
- The `.clang-format` is copied from [Optional 26](https://github.com/beman-project/Optional26/blob/main/.clang-format) & [net29](https://github.com/beman-project/net29/blob/main/.clang-format) with no modification.

Closes: #29 .

PR #46, opened 2024-10-14
feat: add automatic linting suggestion github workflow

linusboehm:
this commit adds a github workflow to automatically run `pre-commit` and suggests changes to a PR, to fix any linting errors. An example error + suggested fix can be seen here: https://github.com/linusboehm/beman_exemplar/pull/2
Suggestions can be accepted + committed from the PR on github.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -7,11 +7,25 @@ on:
 jobs:
   pre-commit:
     runs-on: ubuntu-latest
+    name: pre-commit
+    permissions:
+      contents: read
+      pull-requests: write
+
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-python@v5
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
         with:
-          python-version: '3.13'
+          python-version: 3.13
+
+      - name: Cache pre-commit hooks
+        uses: actions/cache@v2
+        with:
+          path: ~/.cache/pre-commit
+          key: ${{ runner.os }}-precommit-${{ hashFiles('.pre-commit-config.yaml') }}
 
       - name: Get Changed Files
         id: changed-files
@@ -22,3 +36,11 @@ jobs:
       - uses: pre-commit/action@v3.0.1
         with:
           extra_args: --files ${{ steps.changed-files.outputs.all_changed_files }}
+        continue-on-error: true
+
+      - name: suggester / pre-commit
+        uses: reviewdog/action-suggester@v1
+        with:
+          tool_name: pre-commit
+          fail_on_error: true
+          level: warning


wusatosi:
Oh wow, this is fantastic! I have never heard reviewdog before but this is fantastic!

Do this report problems for files that doesn't get auto-formatted? (e.g. markdown files)

camio:
For what this does, there is surprisingly little configuration we'd have to maintain. The reviewdog project looks actively maintained as well. I'm very much in favor of moving forward with this.

#34 is merged now. I think this PR needs to be rebased on top of that to get this PR merged as well.

@linusboehm are there any drawbacks to moving ahead with this PR that we're not seeing? It looks like only upside to me.

.github/workflows/reviewdog.yml
```diff
@@ -0,0 +1,41 @@
...
+          key: ${{ runner.os }}-precommit-${{ hashFiles('.pre-commit-config.yaml') }}
+
+      - name: Install dependencies
+        run: |
+          pip install pre-commit
+
+      - name: Run pre-commit hooks
+        run: |
+          pre-commit run --all-files
+        continue-on-error: true
```
wusatosi, line 34: If a stand-alone workflow is needed, you can replace these with `pre-commit/action@v3.0.1`.

Would recommend linting only on changes, see:
https://github.com/beman-project/exemplar/blob/bd7ad354842e424ca7dd69da7dff5bfaab70e8f6/.github/workflows/pre-commit.yml#L16-L24
for an implementation example.

wusatosi (commented):
Thank you for bringing-in reviewdog, this functionality looks fantastic.

Since #34 is merged, you can simply add the review dog configuration onto `.github/workflows/pre-commit.yml`.

linusboehm:
> @linusboehm are there any drawbacks to moving ahead with this PR that we're not seeing? It looks like only upside to me.

The only time I've tried to use this was in a codebase with the requirement that only changed lines are formatted (instead of whole files) as the codebase didn't have consistent formatting. That seemed to be a bit cumbersome and sometimes lead to formatting changes that obfuscated the underlying code change. Given that beman repos should be formatted consistently according to the format spec, this shouldn't be a problem and I can't think of any other reasons not to give it a try.

I'll rebase and address the review comments soon (@wusatosi thanks for the review).

.github/workflows/reviewdog.yml
```diff
@@ -0,0 +1,41 @@
+name: linter-checks
+on: [pull_request]
+jobs:
+  cpp:
+    permissions:
+      contents: read
+      checks: write
+      issues: write
+      pull-requests: write
```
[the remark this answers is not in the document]
linusboehm: You mean move it to `.github/workflows/pre-commit.yml`? Are the permissions set for that workflow already?

.github/workflows/reviewdog.yml
```diff
@@ -0,0 +1,41 @@
...
+        run: |
+          pre-commit run --all-files
+        continue-on-error: true
+
+      - name: suggester / pre-commit
+        uses: reviewdog/action-suggester@v1
+        with:
+          tool_name: pre-commit
+          fail_on_error: true
+          level: warning
```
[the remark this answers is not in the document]
linusboehm: done

.github/workflows/reviewdog.yml
linusboehm: done

steve-downey (approved):

bretbrownjr (approved):

.github/workflows/pre-commit.yml
```diff
@@ -7,11 +7,25 @@ on:
...
-      - uses: actions/setup-python@v5
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.13
+
+      - name: Cache pre-commit hooks
```
wusatosi, line 24: This comes with pre-commit/action and is actually not needed.

.github/workflows/pre-commit.yml
```diff
@@ -22,3 +36,11 @@ jobs:
       - uses: pre-commit/action@v3.0.1
         with:
           extra_args: --files ${{ steps.changed-files.outputs.all_changed_files }}
+        continue-on-error: true
```
wusatosi, line 39: This is not needed as well

wusatosi (changes requested):
Thanks for the pr!

wusatosi:
Ops didn't realize this is already merged

.github/workflows/pre-commit.yml
linusboehm: How does it set the paths to cache when not explicitly specifying them?

wusatosi: pre-commit/action caches based on the default configuration location. 

Usually these actions come with appropriate caching built-in (not the pip install pre-commit, but oh well).


e.g. see action log:
https://github.com/beman-project/exemplar/actions/runs/11385583999/job/31675772106#step:6:1

See also:
https://github.com/pre-commit/action/blob/main/action.yml

I guess leaving this in is fine, the restored cache by this step will be overwritten/ merged in with that from pre-commit/action step so really no harm is done. ¯⁠\⁠_⁠(⁠ツ⁠)⁠_⁠/⁠¯

.github/workflows/pre-commit.yml
wusatosi: This is in-fact needed. marking this as resovled.

.github/workflows/reviewdog.yml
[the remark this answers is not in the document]
wusatosi: Ohno! Sorry I didn't reply to this comment before merging, all the permissions are needed in `pre-commit.yml`.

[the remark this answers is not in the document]
wusatosi: Re-introduced these permissions at #50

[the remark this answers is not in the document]
wusatosi: Again sorry for not being clear.

PR #48, opened 2024-10-18
Move Build & Test out of docker

wusatosi:
This PR moves the build and test system in CI out of docker and refactors the CI script.

This PR is functionally the same as head (with minor improvement). 

Reduce CI time from ~2 minutes down to ~45 seconds.

Based on Discussion here: https://discourse.bemanproject.org/t/should-we-use-or-remove-docker-from-beman/228

Closes #27

PR #50, opened 2024-10-18
Fix review dog

### LINKED ISSUE #49: Review dog failing in CI due to insufficient permission
Review dog is not generating reviews correctly.

Error report:

```
reviewdog: This GitHub Token doesn't have write permission of Review API [1],
so reviewdog will report results via logging command [2] and create annotations similar to
github-pr-check reporter as a fallback.
```

See: 
image
Link:
https://github.com/beman-project/exemplar/actions/runs/11411570352/job/31756057258?pr=48#step:7:238

CC: @linusboehm

wusatosi:
closes #49 

Reviewdog introduced in #46 is lacking permission to create annotations, see #49 for error logs on #48, commit: 6ec3d4f2aba30d8ff18cda32cec8ea3538b02c60 .

Also 
1. fixed use of deprecated `fail-on-error`, causing warning at: https://github.com/beman-project/exemplar/actions/runs/11411671176/job/31756356511?pr=48#step:7:234
2. Removed unnecessary caching behavior, see: https://github.com/beman-project/exemplar/pull/46#pullrequestreview-2375594753
3. Only trigger review dog on pull request (currently commits will also trigger review dog). 

See example at: https://github.com/wusatosi/exemplar/pull/11

CC: @linusboehm

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -44,5 +44,5 @@ jobs:
         uses: reviewdog/action-suggester@v1
         with:
           tool_name: pre-commit
-          fail_on_error: true
           level: warning
+          reviewdog_flags: "-fail-level=error"
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -10,6 +10,8 @@ jobs:
     name: pre-commit
     permissions:
       contents: read
+      checks: write
+      issues: write
       pull-requests: write
 
     steps:
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -23,12 +23,6 @@ jobs:
         with:
           python-version: 3.13
 
-      - name: Cache pre-commit hooks
-        uses: actions/cache@v2
-        with:
-          path: ~/.cache/pre-commit
-          key: ${{ runner.os }}-precommit-${{ hashFiles('.pre-commit-config.yaml') }}
-
       - name: Get Changed Files
         id: changed-files
         uses: tj-actions/changed-files@v45
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -35,6 +35,7 @@ jobs:
         continue-on-error: true
 
       - name: suggester / pre-commit
+        if: ${{ github.event_name == 'pull_request' }}
         uses: reviewdog/action-suggester@v1
         with:
           tool_name: pre-commit


linusboehm:
Thanks for the change! I'm not quite sure how the permissions work, but I think I recall that the permissions have to be set in the repository settings to allow workflows to comment on a commit.

camio (approved):

wusatosi:
> Thanks for the change! I'm not quite sure how the permissions work, but I think I recall that the permissions have to be set in the repository settings to allow workflows to comment on a commit.

Its interesting that only `pull-requests: write` is not sufficient to leave reviews...
I don't know what's going on as well.

camio:
@wusatosi I attempted to make you a member of the org. Let me know if that doesn't give you authorization to change permissions as needed.

linusboehm:
@camio i don't think that's sufficient. I think someone who's able to change the repository settings needs to change a setting (I can look it up. I think I've done it before). However, this might be a problem for the auto-commenting in general, because I'm not sure changing the setting can be automated. So it might need some manual setup/setting change for every repository that wants to use that feature.

.github/workflows/pre-commit.yml
```diff
@@ -39,8 +35,9 @@ jobs:
         continue-on-error: true
 
       - name: suggester / pre-commit
+        if: ${{ github.event_name == 'pull_request' }}
```
wusatosi, line 38: AH I just realized this breaks linting for comments.

Pre-commit will fail but it has continue on failure set to true for review dog to run, and in commits review dog doesn't run so the check will show up as passed.

I don't think there's anyway to fix this within this job, I have to make a separate job that triggers on commit.

Damn it!

wusatosi:
>> [Simply giving me org member doesn't help]

Let me try to open a pull request to check.

linusboehm:
> > Thanks for the change! I'm not quite sure how the permissions work, but I think I recall that the permissions have to be set in the repository settings to allow workflows to comment on a commit.
> 
> 
> 
> Its interesting that only `pull-requests: write` is not sufficient to leave reviews...
> 
> I don't know what's going on as well.

I tested it with `pull-request: write` only here: https://github.com/linusboehm/beman_exemplar/pull/2
So I don't think the other two permissions that were added are necessary

linusboehm:
this might be the reason: https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#how-the-permissions-are-calculated-for-a-workflow-job

>  if the workflow was triggered by a pull request from a forked repository, and the Send write tokens to workflows from pull requests setting is not selected, the permissions are adjusted to change any write permissions to read only.

wusatosi:
> this might be the reason: https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#how-the-permissions-are-calculated-for-a-workflow-job
> 
> > if the workflow was triggered by a pull request from a forked repository, and the Send write tokens to workflows from pull requests setting is not selected, the permissions are adjusted to change any write permissions to read only.
> 
> @camio so seems like a setting change is required (member of the org is not sufficient, as I can't access the settings)

See #53 , I think linus you are right here.

wusatosi:
I reopened #49 , let's forward the discussion there.

PR #57, opened 2024-10-22
Run TSan and ASan on every C++ version

wusatosi:
Since we will be selectively running tests and examples based on C++ version, we should run TSan and ASan on every C++ version variants in CI. 

This PR updates the matrix to run three flavors of CMake configuration per compiler per C++ version: 
1. Default (No extra argument)
2. TSan (with Thread sanitizer enabled) 
3. ASan (with Address and undefined sanitizer enabled)

This PR also specifies `CMAKE_CXX_STANDARD` for every run to avoid cmake falling back to older C++ versions.

No test case will be removed from this change, total matrix size from build and test increase from 15 to 23.

### DIFF
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -36,43 +36,25 @@ jobs:
             c: gcc
           - cpp: clang++
             c: clang
+        cpp_version: ["17", "20", "23", "26"]
         cmake_args:
-          - description: "Static default"
+          - description: "Default"
             args: ""
-          - description: "Dynamic default"
-            args: "-DBUILD_SHARED_LIBS=on"
-          - description: "static C++17"
-            args: "-DCMAKE_CXX_STANDARD=17"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_tests.yml
```diff
@@ -16,13 +16,14 @@ jobs:
...
       - uses: actions/checkout@v4
-      - name: Setup cmake
-        uses: jwlawson/actions-setup-cmake@v2
+      - name: Setup build environment
+        uses: lukka/get-cmake@latest
         with:
-          cmake-version: '3.25.x'
-      - uses: seanmiddleditch/gha-setup-ninja@v5
+          cmakeVersion: "~3.25.0"
+          ninjaVersion: "^1.11.1"
```
wusatosi, line 26: Just tidying stuff up here. Giving preset test a better name and use the same setup-cmake action as the one below for consistency.

camio (approved):

Issue #59, opened 2024-10-23
The library should have at least one runtime dependency to be really helpful


ClausKlein:
An STATIC, SHARED buildable library should be created, not only a simple header only (INTERFACE).

It should also export only it public interface!

See too https://crascit.com/2019/10/16/cppcon-2019-deep-cmake-for-library-authors/

camio:
This would be great. @ClausKlein would you be willing to contribute something like this?

ClausKlein:
Hhure, I would like to help.

camio:
Thanks! Feel free to create a PR or throw some ideas here. Let me know if there's anything I can do to help.

ClausKlein:
Do you now [WindowsToolchain](https://github.com/MarkSchofield/WindowsToolchain)?
I would like to use them for windows CI build with `ninja`.

This builds quite faster and makes it even more easy to setup a `cmake workflow preset` for windows builds.

bretbrownjr:
Why does this project need a runtime dependency? I think I need to hear some more concrete situations where runtime dependencies are required or at least clearly superior to other options.

Note that this project is intentionally not an educational project to teach CMake [1]. It is intended to assist a reasonably experienced C++ engineer to start developing a standard library proposal.

I would expect the typical proposal would have a fairly modest dependency list, frequently an empty one, but maybe I am too optimistic.

[1] Jason Turner's https://github.com/cpp-best-practices/cmake_template is more in line for those goals.

camio:
> Why does this project need a runtime dependency? I think I need to hear some more concrete situations where runtime dependencies are required or at least clearly superior to other options.

This is a fair question. I would imagine something the linear algebra proposal would have a runtime dependency on blas. The, now defunct, 2d graphics proposal would have a runtime dependency on cairo IIRC.

camio:
> Do you know the [WindowsToolchain](https://github.com/MarkSchofield/WindowsToolchain)?

I was unfamiliar with that. Thanks for pointing it out!

My understanding is that most professional Windows developers use Visual Studio, but I could be out-of-date.

ClausKlein:
> My understanding is that most professional Windows developers use Visual Studio, but I could be out-of-date.

Yes and No, many I know use also `qt-creator, clion`, but not on **CI**.

[...]

PR #68, opened 2024-11-08
Adds vcpkg toolchain into CI

wusatosi:
In complement to #7 , this PR adds CI check that uses the vcpkg tool chain.

Please note that I do not know if CMake is actually using vcpkg, as there's no indication which tool-chain its using (may relate: #59).

Let me know if I should merge #7 first and rebase this pr.


wusatosi:
Looks like this is not ready, (failed due to #71), I will re-open once it's ready.

PR #74, opened 2024-11-13
Apply Beman CMake standard

wusatosi:
Update all reference to `BUILD_TESTING` to `BEMAN_EXEMPLAR_BUILD_TESTING` to conform with beman standard.

PR #75, opened 2024-11-13
Fix CI

wusatosi:
Because every-time I try to update CI there's always problems.

This fixes failed CI due to rushed merge of #74 .

This PR also shortens current job names.

image
image


steve-downey (approved):
LGTM

PR #76, opened 2024-11-14
Introduce CMake toolchain

### LINKED ISSUE #65: clang++: error: unsupported option '-fsanitize=leak' for target 'x86_64-apple-darwin23.6.0'
```bash
bash-5.2$ cmake --workflow --preset gcc-debug  
Executing workflow step 1 of 3: configure preset "gcc-debug"

Preset CMake variables:

  CMAKE_BUILD_TYPE="Debug"
  CMAKE_CXX_COMPILER="g++"
  CMAKE_CXX_FLAGS="-fsanitize=address -fsanitize=pointer-compare -fsanitize=pointer-subtract -fsanitize=leak -fsanitize=undefined"
  CMAKE_CXX_STANDARD="17"

-- The CXX compiler identification is AppleClang 16.0.0.16000026
-- Detecting CXX compiler ABI info - failed
-- Check for working CXX compiler: /usr/bin/g++
-- Check for working CXX compiler: /usr/bin/g++ - broken
CMake Error at /usr/local/Cellar/cmake/3.30.5/share/cmake/Modules/CMakeTestCXXCompiler.cmake:73 (message):
  The C++ compiler

    "/usr/bin/g++"

  is not able to compile a simple test program.

  It fails with the following output:

    Change Dir: '/Users/clausklein/Workspace/cpp/beman-project/exemplar/build/gcc-debug/CMakeFiles/CMakeScratch/TryCompile-h6rdZf'
    
    Run Build Command(s): /usr/local/bin/ninja -v cmTC_abf57
    [1/2] /usr/bin/g++   -fsanitize=address -fsanitize=pointer-compare -fsanitize=pointer-subtract -fsanitize=leak -fsanitize=undefined  -std=gnu++17 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.1.sdk -mmacosx-version-min=14.7 -MD -MT CMakeFiles/cmTC_abf57.dir/testCXXCompiler.cxx.o -MF CMakeFiles/cmTC_abf57.dir/testCXXCompiler.cxx.o.d -o CMakeFiles/cmTC_abf57.dir/testCXXCompiler.cxx.o -c /Users/clausklein/Workspace/cpp/beman-project/exemplar/build/gcc-debug/CMakeFiles/CMakeScratch/TryCompile-h6rdZf/testCXXCompiler.cxx
    FAILED: CMakeFiles/cmTC_abf57.dir/testCXXCompiler.cxx.o 
    /usr/bin/g++   -fsanitize=address -fsanitize=pointer-compare -fsanitize=pointer-subtract -fsanitize=leak -fsanitize=undefined  -std=gnu++17 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.1.sdk -mmacosx-version-min=14.7 -MD -MT CMakeFiles/cmTC_abf57.dir/testCXXCompiler.cxx.o -MF CMakeFiles/cmTC_abf57.dir/testCXXCompiler.cxx.o.d -o CMakeFiles/cmTC_abf57.dir/testCXXCompiler.cxx.o -c /Users/clausklein/Workspace/cpp/beman-project/exemplar/build/gcc-debug/CMakeFiles/CMakeScratch/TryCompile-h6rdZf/testCXXCompiler.cxx
    clang++: error: unsupported option '-fsanitize=leak' for target 'x86_64-apple-darwin23.6.0'
    
    

  

  CMake will not be able to correctly generate this project.
Call Stack (most recent call first):
  CMakeLists.txt:5 (project)


-- Configuring incomplete, errors occurred!
bash-5.2$ 
```

wusatosi:
This PR adopts @bretbrownjr 's suggestion in https://github.com/beman-project/exemplar/pull/44#pullrequestreview-2367123798.

> You can use variables like CMAKE_CXX_FLAGS_Debug_INIT to tune what a Debug or Release build entails, for what it's worth. It's maybe a little nicer to use those variables instead of CMAKE_CXX_FLAGS. But not a huge deal.
>
> https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_FLAGS_CONFIG_INIT.html

This PR introduces toolchain files for supported platforms:
- gcc
- clang
- msvc

Updated CI and preset to use the new toolchain files.

Race with #82


cmake/toolchain.cmake
```diff
@@ -0,0 +1,51 @@
...
+        # Exclude -fsanitize=leak if using GCC on macOS
+        # TODO: Is there a way to detect Apple Clang???
+        if(
+            CMAKE_CXX_COMPILER_ID STREQUAL "GNU"
+            AND CMAKE_SYSTEM_NAME STREQUAL "Darwin"
+        )
+            message(STATUS "Using GCC on macOS; excluding -fsanitize=leak")
+        else()
+            set(SANITIZER_FLAGS "${SANITIZER_FLAGS} -fsanitize=leak")
+        endif()
```
wusatosi, line 32: Is there a better way to detect Apple Clang faking gcc?

steve-downey: I would expect that the compiler identification should figure out if it's really AppleClang even though /usr/bin/g++ is symlinked to clang? 
https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_ID.html

wusatosi: I don't have a macos development environment to test this.

Is AppleClang the identifier?

wusatosi: @steve-downey steve can you explicitly block this PR in review so someone doesn't accidentally come in and merge this?

Recommends message:
We need to verify `CMAKE_CXX_COMPILER_ID` for g++ on macos is `AppleClang`.

steve-downey (changes requested):
"We need to verify CMAKE_CXX_COMPILER_ID for g++ on macos is AppleClang."

Confirm what the compiler identification is for the default false g++ on Darwin is. 

Marking "Request changes" so this doesn't get landed prematurely.

cmake/toolchain.cmake
```diff
@@ -0,0 +1,51 @@
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+set(CMAKE_C_FLAGS_RELEASE_INIT "-O3")
+set(CMAKE_CXX_FLAGS_RELEASE_INIT "-O3")
+
+set(CMAKE_C_FLAGS_RELWITHDEBINFO_INIT "-O3")
+set(CMAKE_CXX_FLAGS_RELWITHDEBINFO_INIT "-O3")
+
+# BEMAN_BUILDSYS_SANITIZER is not a general use option
```
wusatosi, line 9: I am open for more suggestions regarding this option name.

wusatosi:
Maybe we should just pull [Santizer.cmake](https://github.com/cpp-best-practices/cmake_template/blob/main/cmake/Sanitizers.cmake), I foresee our toolchain file just developing into this Sanitzer.cmake.

cmake/toolchain.cmake
ClausKlein:
```bash
iMac:exemplar clausklein$ CXX=g++-14 cmake -S . -B build -G Ninja
-- The CXX compiler identification is GNU 14.2.0
-- Checking whether CXX compiler has -isysroot
-- Checking whether CXX compiler has -isysroot - yes
-- Checking whether CXX compiler supports OSX deployment target flag
-- Checking whether CXX compiler supports OSX deployment target flag - yes
-- Check for working CXX compiler: /usr/local/bin/g++-14 - skipped
-- The C compiler identification is AppleClang 16.0.0.16000026
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Found Python3: /usr/local/Frameworks/Python.framework/Versions/3.13/bin/python3.13 (found version "3.13.0") found components: Interpreter
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE
-- Looking for __cpp_lib_ranges - not found
CMake Warning at examples/CMakeLists.txt:12 (message):
  Missing range support! Skip: identity_as_default_projection


Examples to be built: identity_direct_usage
-- Configuring done (16.8s)
-- Generating done (0.1s)
-- Build files have been written to: /Users/clausklein/Workspace/cpp/beman-project/exemplar/build
iMac:exemplar clausklein$ 
```

wusatosi: Ah! Is the compiler id for AppleClang GNU?!

ClausKlein:
There are build errors on Darwin (native iMac):
```bash
iMac:exemplar clausklein$ cmake --preset gcc-debug --fresh
Preset CMake variables:

  BEMAN_BUILDSYS_SANITIZER="ASan"
  CMAKE_BUILD_TYPE="Debug"
  CMAKE_CXX_COMPILER="g++"
  CMAKE_CXX_STANDARD="20"
  CMAKE_TOOLCHAIN_FILE="cmake/toolchain.cmake"

-- The CXX compiler identification is AppleClang 16.0.0.16000026
-- Check for working CXX compiler: /usr/bin/g++ - skipped
-- The C compiler identification is AppleClang 16.0.0.16000026
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Found Python3: /usr/local/Frameworks/Python.framework/Versions/3.13/bin/python3.13 (found version "3.13.0") found components: Interpreter
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE
-- Looking for __cpp_lib_ranges - found
Examples to be built: identity_direct_usage;identity_as_default_projection
-- Configuring done (2.8s)
-- Generating done (0.0s)
-- Build files have been written to: /Users/clausklein/Workspace/cpp/beman-project/exemplar/build/gcc-debug
iMac:exemplar clausklein$ ninja -C build/gcc-debug -j1
ninja: Entering directory `build/gcc-debug'
[1/12] Building CXX object _deps/googletest-build/googletest/CMakeFiles/gtest.dir/src/gtest-all.cc.o
FAILED: _deps/googletest-build/googletest/CMakeFiles/gtest.dir/src/gtest-all.cc.o 
/usr/bin/g++  -I/Users/clausklein/Workspace/cpp/beman-project/exemplar/build/gcc-debug/_deps/googletest-src/googletest/include -I/Users/clausklein/Workspace/cpp/beman-project/exemplar/build/gcc-debug/_deps/googletest-src/googletest -fsanitize=address -fsanitize=pointer-compare -fsanitize=pointer-subtract -fsanitize=undefined -fsanitize=leak -fsanitize=address -fsanitize=pointer-compare -fsanitize=pointer-subtract -fsanitize=undefined -fsanitize=leak -fsanitize=address -fsanitize=pointer-compare -fsanitize=pointer-subtract -fsanitize=undefined -fsanitize=leak -g -std=c++20 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.1.sdk -mmacosx-version-min=14.7  -DGTEST_HAS_PTHREAD=1 -MD -MT _deps/googletest-build/googletest/CMakeFiles/gtest.dir/src/gtest-all.cc.o -MF _deps/googletest-build/googletest/CMakeFiles/gtest.dir/src/gtest-all.cc.o.d -o _deps/googletest-build/googletest/CMakeFiles/gtest.dir/src/gtest-all.cc.o -c /Users/clausklein/Workspace/cpp/beman-project/exemplar/build/gcc-debug/_deps/googletest-src/googletest/src/gtest-all.cc
clang++: error: unsupported option '-fsanitize=leak' for target 'x86_64-apple-darwin23.6.0'
iMac:exemplar clausklein$ git status
Auf Branch wusatosi/toolchain
nichts zu committen, Arbeitsverzeichnis unverändert
iMac:exemplar clausklein$ 
```

wusatosi:
Ahhh! Looks like it's detecting correctly but not turning off leak sanitizer correctly. 

Thanks for the help CK!

steve-downey:
So, yes:
The CXX compiler identification is AppleClang 16.0.0.16000026

steve-downey:
Not sure if picking the toolchain for Darwin better, or making the generic toolchain smarter would work better, but either should work?

ClausKlein:
Wait, that can't work!
```
iMac:exemplar clausklein$ cmake --preset gcc-debug  --trace-expand --trace-source=toolchain.cmake
Put cmake in trace mode, but with variables expanded.
Put cmake in trace mode, but output only lines of a specified file. Multiple options are allowed.
Preset CMake variables:

  BEMAN_BUILDSYS_SANITIZER="ASan"
  CMAKE_BUILD_TYPE="Debug"
  CMAKE_CXX_COMPILER="g++"
  CMAKE_CXX_STANDARD="20"
  CMAKE_TOOLCHAIN_FILE="cmake/toolchain.cmake"

/Users/clausklein/Workspace/cpp/beman-project/exemplar/cmake/toolchain.cmake(3):  set(CMAKE_C_FLAGS_RELEASE_INIT -O3 )
/Users/clausklein/Workspace/cpp/beman-project/exemplar/cmake/toolchain.cmake(4):  set(CMAKE_CXX_FLAGS_RELEASE_INIT -O3 )
/Users/clausklein/Workspace/cpp/beman-project/exemplar/cmake/toolchain.cmake(6):  set(CMAKE_C_FLAGS_RELWITHDEBINFO_INIT -O3 )
/Users/clausklein/Workspace/cpp/beman-project/exemplar/cmake/toolchain.cmake(7):  set(CMAKE_CXX_FLAGS_RELWITHDEBINFO_INIT -O3 )
/Users/clausklein/Workspace/cpp/beman-project/exemplar/cmake/toolchain.cmake(16):  if(DEFINED BEMAN_BUILDSYS_SANITIZER )
/Users/clausklein/Workspace/cpp/beman-project/exemplar/cmake/toolchain.cmake(17):  if(BEMAN_BUILDSYS_SANITIZER STREQUAL ASan )
/Users/clausklein/Workspace/cpp/beman-project/exemplar/cmake/toolchain.cmake(19):  set(SANITIZER_FLAGS -fsanitize=address -fsanitize=pointer-compare -fsanitize=pointer-subtract -fsanitize=undefined )
/Users/clausklein/Workspace/cpp/beman-project/exemplar/cmake/toolchain.cmake(24):  if(NOT CMAKE_CXX_COMPILER_ID )
/Users/clausklein/Workspace/cpp/beman-project/exemplar/cmake/toolchain.cmake(25):  message(WARNING toolchain is used before CMAKE_CXX_COMPILER_ID was set! )
CMake Warning at cmake/toolchain.cmake:25 (message):
  toolchain is used before CMAKE_CXX_COMPILER_ID was set!
Call Stack (most recent call first):
  build/gcc-debug/CMakeFiles/3.31.0-dirty/CMakeSystem.cmake:6 (include)
  CMakeLists.txt:5 (project)


/Users/clausklein/Workspace/cpp/beman-project/exemplar/cmake/toolchain.cmake(27):  if(APPLE )
/Users/clausklein/Workspace/cpp/beman-project/exemplar/cmake/toolchain.cmake(28):  message(STATUS Using GCC on macOS; excluding -fsanitize=leak )
-- Using GCC on macOS; excluding -fsanitize=leak
/Users/clausklein/Workspace/cpp/beman-project/exemplar/cmake/toolchain.cmake(44):  list(APPEND CMAKE_C_FLAGS_DEBUG_INIT -fsanitize=address -fsanitize=pointer-compare -fsanitize=pointer-subtract -fsanitize=undefined )
/Users/clausklein/Workspace/cpp/beman-project/exemplar/cmake/toolchain.cmake(45):  list(APPEND CMAKE_CXX_FLAGS_DEBUG_INIT -fsanitize=address -fsanitize=pointer-compare -fsanitize=pointer-subtract -fsanitize=undefined )
-- The CXX compiler identification is AppleClang 16.0.0.16000026
/Users/clausklein/Workspace/cpp/beman-project/exemplar/cmake/toolchain.cmake(3):  set(CMAKE_C_FLAGS_RELEASE_INIT -O3 )
/Users/clausklein/Workspace/cpp/beman-project/exemplar/cmake/toolchain.cmake(4):  set(CMAKE_CXX_FLAGS_RELEASE_INIT -O3 )
/Users/clausklein/Workspace/cpp/beman-project/exemplar/cmake/toolchain.cmake(6):  set(CMAKE_C_FLAGS_RELWITHDEBINFO_INIT -O3 )
/Users/clausklein/Workspace/cpp/beman-project/exemplar/cmake/toolchain.cmake(7):  set(CMAKE_CXX_FLAGS_RELWITHDEBINFO_INIT -O3 )
/Users/clausklein/Workspace/cpp/beman-project/exemplar/cmake/toolchain.cmake(16):  if(DEFINED BEMAN_BUILDSYS_SANITIZER )
-- Check for working CXX compiler: /usr/bin/g++ - skipped
Examples to be built: identity_direct_usage;identity_as_default_projection
-- Configuring done (2.4s)
```

cmake/toolchain.cmake
ClausKlein: I had to changed this:
```diff
diff --git a/cmake/toolchain.cmake b/cmake/toolchain.cmake
index b3cd128..450e8a9 100644
--- a/cmake/toolchain.cmake
+++ b/cmake/toolchain.cmake
@@ -21,10 +21,13 @@ if(DEFINED BEMAN_BUILDSYS_SANITIZER)
         )
 
         # Exclude -fsanitize=leak on Apple Clang (gcc on macos) as it is not supported.
-        if(CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
+        if(NOT CMAKE_CXX_COMPILER_ID)
+            message(WARNING "toolchain is used before CMAKE_CXX_COMPILER_ID was set!")
+        endif()
+        if(APPLE) # AND ((${CMAKE_CXX_COMPILER_ID} MATCHES "Clang") OR (${CMAKE_CXX_COMPILER_ID} MATCHES "GNU")))
             message(STATUS "Using GCC on macOS; excluding -fsanitize=leak")
         else()
-            set(SANITIZER_FLAGS "${SANITIZER_FLAGS} -fsanitize=leak")
+            list(APPEND SANITIZER_FLAGS "-fsanitize=leak")
         endif()
     elseif(BEMAN_BUILDSYS_SANITIZER STREQUAL "TSan")
         # Basic ASan flags
@@ -38,10 +41,6 @@ if(DEFINED BEMAN_BUILDSYS_SANITIZER)
         )
     endif()
 
-    set(CMAKE_C_FLAGS_DEBUG_INIT
-        "${CMAKE_C_FLAGS_DEBUG_INIT} ${SANITIZER_FLAGS}"
-    )
-    set(CMAKE_CXX_FLAGS_DEBUG_INIT
-        "${CMAKE_C_FLAGS_DEBUG_INIT} ${SANITIZER_FLAGS}"
-    )
+    list(APPEND CMAKE_C_FLAGS_DEBUG_INIT "${SANITIZER_FLAGS}")
+    list(APPEND CMAKE_CXX_FLAGS_DEBUG_INIT "${SANITIZER_FLAGS}")
 endif()
```

wusatosi:
> Not sure if picking the toolchain for Darwin better, or making the generic toolchain smarter would work better, but either should work? 

I am leaning on having separate tool chain files and having like a central tool chain dispatch logic based on compiler and platform. 

But then I realized there's not that much variance in building across platforms and compilers, at least in exemplar, to warrant separate files.

cmake/toolchain.cmake
ClausKlein: And the string, list handling in cmake is petty!
```bash
ninja -C build/gcc-debug -j1
ninja: Entering directory `build/gcc-debug'
[1/12] Building CXX object _deps/googletest-build/googletest/CMakeFiles/gtest.dir/src/gtest-all.cc.o
FAILED: _deps/googletest-build/googletest/CMakeFiles/gtest.dir/src/gtest-all.cc.o 
/usr/bin/g++  -I/Users/clausklein/Workspace/cpp/beman-project/exemplar/build/gcc-debug/_deps/googletest-src/googletest/include -I/Users/clausklein/Workspace/cpp/beman-project/exemplar/build/gcc-debug/_deps/googletest-src/googletest -fsanitize=address -fsanitize=pointer-compare -fsanitize=pointer-subtract -fsanitize=undefined;-fsanitize=address -fsanitize=pointer-compare -fsanitize=pointer-subtract -fsanitize=undefined -g -std=c++20 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.1.sdk -mmacosx-version-min=14.7  -DGTEST_HAS_PTHREAD=1 -MD -MT _deps/googletest-build/googletest/CMakeFiles/gtest.dir/src/gtest-all.cc.o -MF _deps/googletest-build/googletest/CMakeFiles/gtest.dir/src/gtest-all.cc.o.d -o _deps/googletest-build/googletest/CMakeFiles/gtest.dir/src/gtest-all.cc.o -c /Users/clausklein/Workspace/cpp/beman-project/exemplar/build/gcc-debug/_deps/googletest-src/googletest/src/gtest-all.cc
clang++: error: no input files
/bin/sh: -fsanitize=address: command not found
```

ClausKlein:
> But then I realized there's not that much variance in building across platforms and compilers, at least in exemplar, to warrant separate files.

That is why often I use the [project_options](https://github.com/aminya/project_options)
see i.e.: https://github.com/aminya/project_options/blob/main/src/Sanitizers.cmake

wusatosi:
> That is why often I use the [project_options](https://github.com/aminya/project_options) see i.e.: https://github.com/aminya/project_options/blob/main/src/Sanitizers.cmake

That's project looks fantastic!

I can bring this up in weekly sync and see if we want to use this.
Clang-tidy support, coverage, doxygen (and potentially vcpkg) are improvement features we would love to have, it would be fantastic if we can get all these done with this dependency.

cmake/toolchain.cmake
wusatosi: wait why is compiler identification GNU here but AppleClang below?

wusatosi:
Ah I think tool chain file is executed before `project()`, so `CMAKE_CXX_COMPILER_ID` maybe unset?

ClausKlein:
> Ah I think tool chain file is executed before `project()`, so `CMAKE_CXX_COMPILER_ID` maybe unset?

Yes, and call **N** times for each **LANGUAGES** enabled in `project`.

wusatosi:
Why is GitHub Actions not triggering again...

wusatosi:
So I decide to not implement a toolchain file instead adding a generic `apply_sanitizer` util cmake script under the `cmake` directory. I will update the PR description accordingly.

I added a macos run of `gcc-debug` to confirm we are building correctly.

@steve-downey you may review and unblock this PR

ClausKlein:
**FYI**: I have added this as demo hers:
https://github.com/beman-project/execution26/pull/80

wusatosi:
nudging @steve-downey @neatudarius @camio to review as #80 maybe improved if this is merged.

camio (changes requested):
My understanding is that this PR is an attempt to provide MacOS compatibility with our GCC preset by attempting to detect the appropriate sanitizer flags on a per-compiler basis.

I do not think this approach is consistent with the principal we've used elsewhere: the invoker of CMake is responsible for setting the compiler and its flags.

MacOS support can be achieved by adding new presets that are specific to the AppleClang compiler which include the appropriate sanitizer flags.

ClausKlein:
> MacOS support can be achieved by adding new presets that are specific to the AppleClang compiler which include the appropriate sanitizer flags.

The `gcc-14` and `AppleClang` are currently not so important on `macOX`.

But `clang-19` must be tested on **APPLE**.!
This is the only available compiler on `unix` with usable `CXX_MODULE` support.

wusatosi:
> My understanding is that this PR is an attempt to provide MacOS compatibility with our GCC preset by attempting to detect the appropriate sanitizer flags on a per-compiler basis.
> 
> I do not think this approach is consistent with the principal we've used elsewhere: the invoker of CMake is responsible for setting the compiler and its flags.
> 
> MacOS support can be achieved by adding new presets that are specific to the AppleClang compiler which include the appropriate sanitizer flags.

The initial intention is to reduce the amount of config in preset and avoid duplicate code in CI and preset.

If our direction is to have a minimal CMake script (which would be consistent with dec 2nd's meeting), this script would not be applicable.

I agree macos support via apple clang specific config would be a better choice, as if this PR is used, apple clang silently passes.

Closing this right now.

camio:
> > MacOS support can be achieved by adding new presets that are specific to the AppleClang compiler which include the appropriate sanitizer flags.
> 
> The `gcc-14` and `AppleClang` are currently not so important on `macOX`.

I agree that gcc-14 isn't important, but AppleClang, being used by 99% of developers on MacOS, is critically important for a preset.

> But `clang-19` must be tested on **APPLE**.! This is the only available compiler on `unix` with usable `CXX_MODULE` support.

Nothing here prevents people from testing clang-19 builds on MacOS, but being a very niche use case, I don't think it belongs in either CI or our presets file.

camio:
> @camio I just realized this is still needed for CI, I can move it under the CI script.
> 
> Basically we would ideally want a CI script of this following matrix:
> 
> compiler: [GNU, clang, vs code, MSVC]

What do you mean here? vs code isn't a compiler.

> config: [default, sanitizer-set-thread, sanitizer-set-Address]
> 
> With this CMake script, we can let GitHub do it's mix and match without manually entering all the permutations. But if we do it from a purely command line interface, the CI script would be non-extendable.

Since each compiler has its own set of flags and options, attempting to abstract the idea of a "configuration" that applies to some compilers, but not others, is going to be difficult to maintain. If we did want such a thing, I still argue it doesn't belong in our CMakeLists.txt files as the decider of which C++ flags to use is the _invoker_ of these scripts, be that a preset, toolchain, or command-line invocation.

I don't think I fully understand what problem you're attempting to solve. What, at the end of the day, would you like our CI jobs to be?

wusatosi:
Sorry about my unclear communication, I am very sleep deprived from finals.

> What do you mean here? vs code isn't a compiler.

Ah, I mean xcode here.

**This is not a simple compatibility layer for apple-clang alone, this is a compatibility layer for all compilers that we support & use in CI and preset.** It's just apple-clang is the only one that someone here tested.

Basically, all the compiler have different support for sanitizers, but these sanitizers could be generally grouped into two groups (tsan and asan), it would be clean to declare on GitHub Actions that I want to generate the permutation of [santizier set: [clang...], c++ version set: [17, 20, ...], sanitizer set (cmake args): [default, tsan-set, asan-set]] (e.g. combination: [clang, 17, asan-set]), and have a subsequent script that determines what the ASan set mean for each compilers (e.g. clang's asan set expands to `-fsantizer=address -fsantizer=undefined ...`, MSVC's asan set expands to `/fsanitize=address /Zi`).

Otherwise we will have to include all the combinations individually, which would lead to hell like:
https://github.com/bemanproject/exemplar/blob/d9145b52c9072a569d747547edfc7e03a1494809/.github/workflows/ci_tests.yml#L30-L44

And this is when there isn't a "different compiler need different command" scenario yet.

You can already see the pain [starting off in the MSVC CI PR](https://github.com/bemanproject/exemplar/pull/80/files#diff-bf607a21ebbd18b3f436cc9fac6937d9c8839c33ae022cb29c2bddd873f470ebR70-R89). I have to include MSVC builds as extras (not part of the permutation), as a result, it only runs ASan on C++17, otherwise the matrix would be:

```
      matrix:
        platform:
          - description: "ubuntu gcc"
            cpp: g++
            c: gcc
            os: ubuntu-latest
          - description: "ubuntu clang"
            cpp: clang++
            c: clang
            os: ubuntu-latest
        cpp_version: [17, 20, 23, 26]
        cmake_args:
          - description: "Default"
            args: ""
          - description: "TSan"
            args: "-DCMAKE_CXX_FLAGS=-fsanitize=thread"
          - description: "ASan"
            args: "-DCMAKE_CXX_FLAGS='-fsanitize=address -fsanitize=undefined'"
        include:
          - platform:
              description: "windows MSVC"
              cpp: cl
              c: cl
              os: windows-latest
            cpp_version: 17
            cmake_args:
              description: "ASan"
              # Debug infomation needed to avoid cl: C5072
              # https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-c5072?view=msvc-170
              args: "-DCMAKE_CXX_FLAGS='/fsanitize=address /Zi'"
          - platform:
              description: "windows MSVC"
              cpp: cl
              c: cl
              os: windows-latest
            cpp_version: 20
            cmake_args:
              description: "ASan"
              # Debug infomation needed to avoid cl: C5072
              # https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-c5072?view=msvc-170
              args: "-DCMAKE_CXX_FLAGS='/fsanitize=address /Zi'"
          - platform:
              description: "windows MSVC"
              cpp: cl
              c: cl
              os: windows-latest
            cpp_version: 23
            cmake_args:
              description: "ASan"
              # Debug infomation needed to avoid cl: C5072
              # https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-c5072?view=msvc-170
              args: "-DCMAKE_CXX_FLAGS='/fsanitize=address /Zi'"
          - platform:
              description: "windows MSVC"
              cpp: cl
              c: cl
              os: windows-latest
            cpp_version: 26
            cmake_args:
              description: "ASan"
              # Debug infomation needed to avoid cl: C5072
              # https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-c5072?view=msvc-170
              args: "-DCMAKE_CXX_FLAGS='/fsanitize=address /Zi'"
```

For MSVC-ASan to cover C++ 17..26. This is **64** lines of CI code with high level of repetition to generate this matrix. Where **43** lines of code is added simply because MSVC use a different options interface and have a very restricted ASan support. This would be needed for apple clang as well once we add apple clang tests to CI because the set of sanitizer it supports is different (the LOC added may be doubled because apple-clang doesn't support TSan). These combination would get worse if we want to add C++14 support as well.

If we have a utility like what this PR provides the matrix would be simplified to:
```
      matrix:
        platform:
          - description: "ubuntu gcc"
            cpp: g++
            c: gcc
            os: ubuntu-latest
          - description: "ubuntu clang"
            cpp: clang++
            c: clang
            os: ubuntu-latest
          - description: "windows MSVC"
            cpp: cl
            c: cl
            os: windows-latest
        cpp_version: [17, 20, 23, 26]
        cmake_args:
          - description: "Default"
            args: ""
          - description: "TSan"
            args: "-DBEMAN_BUILDSYS_SANITIZER=TSan"
          - description: "ASan"
            args: "-DBEMAN_BUILDSYS_SANITIZER=ASan"
    exclude:
        # MSVC CL does not support TSan
        - platform:
            cpp: cl
          cmake_args:
           description: "TSan"
```

Where the only addition needed is:
```
          - description: "windows MSVC"
            cpp: cl
            c: cl
            os: windows-latest
```

and

```
    exclude:
        # MSVC CL does not support TSan
        - platform:
            cpp: cl
          cmake_args:
           description: "TSan"
```

If we create a python script/ CI step that runs a bash script to generate what TSan/ ASan is, then we are essentially rewriting this PR in another language. But writing this as a CMake utility has an advantage as:

This problem not only applies for the CI (if not especially applies to CI), but also presets. By adding a new compiler/ toolchain, we have to specify what ASan is as we want `xxx-debug` to match the ASan sanitizer set on CI. The problem there is not that we need to write an extra line of code, its that we might have two sets of definition on what ASan is across CI and preset. 

If someone updates CI/ preset without updating the other (maybe because the hundreds of lines of code in CI), we have a potential "it works on my machine but not on CI" issue or vice-versa. This would be potentially really hard to debug.

I see the need to have a simple CMake infrastructure, and I think this utlility don't need to be included in the main CMake script. But I don't think the tradeoff of "having minimal CMake script and offload everything to CI" is reasonable in this specific instance. We have to deal with this complexity if we want to run sanitizers in CI, if we want to use the a declarative way to generate the matrix we will have to write a specific script to deal with sanitizer support, CMake script like what's included in this PR would be the most ideal way to implement.

camio:
@wusatosi, thanks for this explanation. I think I understand what you're going for now.

> > What do you mean here? vs code isn't a compiler.
>
> Ah, I mean xcode here.

XCode is also an IDE. I think when you say XCode, you're trying to refer to the native MacOS compiler, which is called AppleClang.

> sanitizers could be generally grouped into two groups (tsan and asan)

This grouping is not well-defined. You mention that ASan is enabled on GCC with the `-fsanitize=address -fsanitize=undefined` flags, but that is enabling two distinct sanitizers: address sanitizer, and undefined behavior sanitizer.

What if the grouping has a different shape? GCC's thread sanitizer is unique in that it may not be used with other sanitizers. Say we create a flag configuration named `RUNTIME_INSTRUMENTATION` that is defined to be the recommended intersection of compatible runtime instrumentation flags for a given compiler. We could have another flag configuration called `ENFORCED_WARNINGS` which would be the recommended intersection of warning and error-on-warning flags for a given compiler. For TSan-like instrumentation we could also create a `MULTITHREAD_RUNTIME_INSTRUMENTATION`, but I don't think we have a use case for that just yet.

The next question is where to put these flag groupings. They don't belong in our `CMakeLists.txt` because the `cmake` invoker is responsible for providing compiler flags. They also don't belong in a flattened CI matrix for the good reasons you pointed out.

What if we create per-compiler toolchains that make use of a "BEMAN_FLAG_SET" variable? The toolchain for AppleClang could look something like this:

```CMake
set(CMAKE_CXX_COMPILER /usr/bin/clang++)
set(CMAKE_C_COMPILER /usr/bin/clang)
if( "RUNTIME_INSTRUMENTATION" IN_LIST BEMAN_FLAG_SET )
    list(APPEND CMAKE_CXX_FLAGS "-fsanitize=address" "-fsanitize=pointer-compare" "-fsanitize=pointer-subtract" "-fsanitize=leak" "-fsanitize=undefined")
    # TODO: Error out if "MULTITHREAD_RUNTIME_INSTRUMENTATION" is also within BEMAN_FLAG_SET
endif()
if( "ENFORCED_WARNINGS" IN_LIST BEMAN_FLAG_SET )
    list(APPEND CMAKE_CXX_FLAGS "-Wall" "-Wextra" "-Werror")
endif()
# ...
```

A similar file is created for each of our supported compilers. CMake can then be invoked with options like `-DCMAKE_TOOLCHAIN_FILE=/path/to/BemanAppleClangToolchain.cmake -DBEMAN_FLAG_SET=RUNTIME_INSTRUMENTATION`.

I believe this would enable us to have a simple CI matrix. These toolchains can potentially live in a separate repository that is shared by the CI specifications for all libraries. (Aside: If our CI configurations check out a particular commit id of the toolchain repository, it would facilitate a gradual migration to toolchain improvements).

A drawback of this approach is that our presets will repeat some of the information in our CI toolchain files. This is an acceptable tradeoff IMO to minimize the complexity of Beman build files.

steve-downey:
Clang's memory sanitizer also doesn't play well with other sanitizers. It also has the unfortunate property of needing to be globally applied, that is, all the libraries that touch memory need to be built with msan, just like tsan, and for essentially the same reason.

wusatosi:
> > sanitizers could be generally grouped into two groups (tsan and asan)
> 
> This grouping is not well-defined. You mention that ASan is enabled on GCC with the `-fsanitize=address -fsanitize=undefined` flags, but that is enabling two distinct sanitizers: address sanitizer, and undefined behavior sanitizer.

By this grouping I mean to have "the minimum number of distinct sanitizer set that allow us to cover all the sanitizers". This was the original design goal for sanitizers on CI.  Since thread sanitizer usually conflict with other sanitizers, this grouping comes down to TSan (thread santizier) and ASan (basically all other santiziers with the main interest being address sanitizer (also this is a good synonyms for All-other-santiziers?)).

The documentation of these set is included in this PR.
https://github.com/bemanproject/exemplar/blob/55c966c13234f972d2892f2de85f49997717d04f/cmake/apply_santizers.cmake#L3-L6

> What if the grouping has a different shape? GCC's thread sanitizer is unique in that it may not be used with other sanitizers. Say we create a flag configuration named `RUNTIME_INSTRUMENTATION` that is defined to be the recommended intersection of compatible runtime instrumentation flags for a given compiler. We could have another flag configuration called `ENFORCED_WARNINGS` which would be the recommended intersection of warning and error-on-warning flags for a given compiler. For TSan-like instrumentation we could also create a `MULTITHREAD_RUNTIME_INSTRUMENTATION`, but I don't think we have a use case for that just yet.

I want to point to the documentation I included in the PR again, the intention for sanitizer group is only to accommodating the fact that we cannot enable all sanitizers all at once. This is not an abstraction layer for general instrumentation based tooling. This is to optimize the common case that we want all the compiler sanitizers we can have, not to perfectly define "threading instrumentation", IMO this would be too much clutter for downstream.

Let's say one day Address santizier doesn't work with undefined santiziers anymore, at that point (and my intention is to only at that point) we can include an extra set.

In a sense, maybe I should just call all the santiziers groups "group a", "group b" instead of "ASan", "TSan" to avoid confusion.

> The next question is where to put these flag groupings. They don't belong in our `CMakeLists.txt` because the `cmake` invoker is responsible for providing compiler flags. They also don't belong in a flattened CI matrix for the good reasons you pointed out.
> 
> What if we create per-compiler toolchains that make use of a "BEMAN_FLAG_SET" variable? The toolchain for AppleClang could look something like this:
> 
> ```cmake
> set(CMAKE_CXX_COMPILER /usr/bin/clang++)
> set(CMAKE_C_COMPILER /usr/bin/clang)
> if( "RUNTIME_INSTRUMENTATION" IN_LIST BEMAN_FLAG_SET )
>     list(APPEND CMAKE_CXX_FLAGS "-fsanitize=address" "-fsanitize=pointer-compare" "-fsanitize=pointer-subtract" "-fsanitize=leak" "-fsanitize=undefined")
>     # TODO: Error out if "MULTITHREAD_RUNTIME_INSTRUMENTATION" is also within BEMAN_FLAG_SET
> endif()
> if( "ENFORCED_WARNINGS" IN_LIST BEMAN_FLAG_SET )
>     list(APPEND CMAKE_CXX_FLAGS "-Wall" "-Wextra" "-Werror")
> endif()
> # ...
> ```
> 
> A similar file is created for each of our supported compilers. CMake can then be invoked with options like `-DCMAKE_TOOLCHAIN_FILE=/path/to/BemanAppleClangToolchain.cmake -DBEMAN_FLAG_SET=RUNTIME_INSTRUMENTATION`.
> 
> I believe this would enable us to have a simple CI matrix. These toolchains can potentially live in a separate repository that is shared by the CI specifications for all libraries. (Aside: If our CI configurations check out a particular commit id of the toolchain repository, it would facilitate a gradual migration to toolchain improvements).
> 
> A drawback of this approach is that our presets will repeat some of the information in our CI toolchain files. This is an acceptable tradeoff IMO to minimize the complexity of Beman build files.

Per-compiler Toolchain was the original intention behind this PR (or what was originally planned to come after this).
Every compiler gets their own toolchain file that specifies sanitizer set A, sanitizer set B.
But I think this goes against the minimal cmake philosophy we are going with here, especially when the code snippet above will likely be the only code needed for each toolchain, it kinda screams "I can be simplified". 

In the simplistic world all the options generically works across all the compilers, I don't see enough variations across compilers to be used that warrants we creating various toolchain files in exemplar.

I want to again point out that the intended audience for this utility here is only CI and preset. This is not intended to be public facing, I think it is obvious any specialization that will be needed downstream (aside from turning off some sanitizer) will need the contributor to create special infrastructure for their respective repo. But I believe the common case for projects are the default sanitizers are good enough without extra per-compiler configuration that deviates from the common `-fsantizer=xxx`.

https://github.com/bemanproject/exemplar/blob/55c966c13234f972d2892f2de85f49997717d04f/CMakeLists.txt#L32-L33

steve-downey:
An important reason for keeping flags out of the core CI is to be buildable by package managers, which need to be in charge of the compilation and compilers being used in order to do their jobs. Different compilers with the same flags are sometimes as incompatible as the same compiler with different flags. If the package manager can supply its toolchain files and rely on us not messing with the request, it can mostly work out of the box.

camio:
> Per-compiler Toolchain was the original intention behind this PR (or what was originally planned to come after this). Every compiler gets their own toolchain file that specifies sanitizer set A, sanitizer set B. But I think this goes against the minimal cmake philosophy we are going with here, especially when the code snippet above will likely be the only code needed for each toolchain, it kinda screams "I can be simplified".

This PR proposes to add complexity to the top-level CMakeLists.txt and platform-specific flag selection. What I am proposing involves no changes to the top-level CMakeLists.txt file and no platform-specific flag selection. All complexity is moved to CI and the toolchains it uses. I disagree that my proposal goes against the minimal CMake approach.

ClausKlein:
> This PR proposes to add complexity to the top-level CMakeLists.txt and platform-specific flag selection. What I am proposing involves no changes to the top-level CMakeLists.txt file and no platform-specific flag selection. All complexity is moved to CI and the toolchains it uses. I disagree that my proposal goes against the minimal CMake approach.

You may use a simpler solution: https://github.com/bemanproject/optional26/pull/85#discussion_r1878818770

camio:
> > This PR proposes to add complexity to the top-level CMakeLists.txt and platform-specific flag selection. What I am proposing involves no changes to the top-level CMakeLists.txt file and no platform-specific flag selection. All complexity is moved to CI and the toolchains it uses. I disagree that my proposal goes against the minimal CMake approach.
> 
> You may use a simpler solution like this: [bemanproject/optional26#85 (comment)](https://github.com/bemanproject/optional26/pull/85#discussion_r1878818770)

If that snippet you pointed to was included in the toolchain files and _not included in the top-level CMakeLists.txt file_, my concerns would be addressed.

ClausKlein:
> If that snippet you pointed to was included in the toolchain files and _not included in the top-level CMakeLists.txt file_, my concerns would be addressed.

The toolchain file are not realy needed.
They set `CXX_FLAGS `  and `LDFAGS ` directly, which is not recommented!

The compiler may set on environment `CXX=clang++-19` cmake ....

The `sanitizer options` may be set in the `cmake configure presets` as needed.

wusatosi:
> This PR proposes to add complexity to the top-level CMakeLists.txt and platform-specific flag selection. What I am proposing involves no changes to the top-level CMakeLists.txt file and no platform-specific flag selection. All complexity is moved to CI and the toolchains it uses. I disagree that my proposal goes against the minimal CMake approach.

Okay I can try to go back to implement toolchain file

wusatosi:
@camio I implemented this using toolchain files, is this more what you are looking for?

cmake/gnu-toolchain.cmake
```diff
@@ -0,0 +1,17 @@
+set(CMAKE_C_COMPILER gcc)
+set(CMAKE_CXX_COMPILER g++)
+
+if(BEMAN_BUILDSYS_SANITIZER STREQUAL "ASan")
```
steve-downey, line 4: Modeling sanitizers as a build type works better, either that or an entirely distinct toolchain so Thread and Memory can be uniformly applied to all packages. If everything in an address space aren't using msan or tsan the reports they provide are broken, so you have to rebuild and relink the whole world consistently. 
UB sanitizer and address, don't suffer the same problems. 

So something like (not tested!):
```cmake
set(CMAKE_CXX_FLAGS_ASAN
    "${CMAKE_CXX_FLAGS_RELWITHDEBINFO} -fsanitize=address,undefined,leak"
    CACHE STRING
    "C++ ASAN Flags"
    FORCE
)
```

Also at -O0 there's often no undefined behavior emitted for the sanitizer to see, for the same reasons that debug builds seg fault less often.

steve-downey: (optional26 doesn't use the _INIT variables because it's copied from ancient sources before that rule was clarified. Above should be using the *_INIT vars)

cmake/gnu-toolchain.cmake
```diff
@@ -0,0 +1,17 @@
+set(CMAKE_C_COMPILER gcc)
```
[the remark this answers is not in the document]
wusatosi: 👌

cmake/gnu-toolchain.cmake
wusatosi: Thanks for the write-up.

Separate ASAN as build targets seems a bit like overkill for the exemplars use case. 

The design goal here isn't to have a full fledged instrumentation based analysis build system but just a quick hand for "enable all flags for sanitizers". 

Given there's no dependency for exemplar, and the current recommendation for dependency management is to build with dependency's source code instead of including the dependent library at link time. I don't think there's value in complications here.

cmake/gnu-toolchain.cmake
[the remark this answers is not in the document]
wusatosi: Adopted

cmake/gnu-toolchain.cmake
wusatosi: Ah I see where u r coming from, I didn't think about exemplar as a dependency for other libraries (being a dependent) and were only commenting on its use of dependency and a standalone development library / CI test target. 

I think you are right, there should be an ASAN target to produce an ASAN enabled library so someone could link us as a dependency to use. I get what you are talking about. But I think this is more of a package/ export issue, outside of scope for this PR for now and to be honest outside of my skill tree for now.

Again again again, the main motivation here is just to simply CI/ workflow.

Honestly I am tentatively waiting for someone to implement package export, do a quick write up, evaluate it and yonk it over (just like code coverage).

Could we delegate this suggestion to another PR? Let me know if I should add something/ structure this tool chain in anticipation of this feature.

steve-downey:
Figuring out better ergonomics for handling sanitizers (and fuzzers, coverage, and the rest of the laundry list) can be ongoing work. Getting sanitizers in CI is an immediate improvement. 

I would base the sanitizers on the release or relwithdebinfo profile in CI, as debug tends to not exercise any of the runtime problems that the sanitizers detect.

cmake/gnu-toolchain.cmake
camio: > Modeling sanitizers as a build type works better, either that or an entirely distinct toolchain so Thread and Memory can be uniformly applied to all packages.

How so? This isn't obvious for me.

cmake/llvm-toolchain.cmake
```diff
@@ -0,0 +1,21 @@
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+include_guard(GLOBAL)
+
+set(CMAKE_C_COMPILER clang)
+set(CMAKE_CXX_COMPILER clang++)
+
+if(BEMAN_BUILDSYS_SANITIZER STREQUAL "ASan")
```
camio, line 8: ASan already has a well-known meaning in the community: address sanitizer. Using the same word to mean "all sanitizers" is confusing.

What about "MaxSan" (maximum sanitizers) which would mean getting as many sanitizers enabled as possible.

cmake/llvm-toolchain.cmake
```diff
@@ -0,0 +1,21 @@
...
+    set(CMAKE_CXX_FLAGS_DEBUG_INIT
+        "-fsanitize=address -fsanitize=leak -fsanitize=pointer-compare -fsanitize=pointer-subtract -fsanitize=undefined"
+    )
+    set(CMAKE_C_FLAGS_DEBUG_INIT
+        "-fsanitize=address -fsanitize=leak -fsanitize=pointer-compare -fsanitize=pointer-subtract -fsanitize=undefined"
+    )
```
camio, line 14: Would this not also apply to release builds?

cmake/msvc-toolchain.cmake
```diff
@@ -0,0 +1,19 @@
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+include_guard(GLOBAL)
+
+set(CMAKE_C_COMPILER cl)
+set(CMAKE_CXX_COMPILER cl)
+
+if(BEMAN_BUILDSYS_SANITIZER STREQUAL "ASan")
+    set(CMAKE_CXX_FLAGS_DEBUG_INIT "/fsanitize=address /Zi")
+    set(CMAKE_C_FLAGS_DEBUG_INIT "/fsanitize=address /Zi")
```
camio, line 10: Is `/Zi` is related to ASan? Should it be enabled in all build modes?

cmake/llvm-toolchain.cmake
```diff
@@ -0,0 +1,21 @@
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```
camio, line 1: This toolchain needs file-level documentation explaining its purpose and the semantics of its parameters (viz. BEMAN_BUILDSYS_SANITIZER)

cmake/llvm-toolchain.cmake
```diff
@@ -0,0 +1,21 @@
...
+
+include_guard(GLOBAL)
```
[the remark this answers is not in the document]
wusatosi: I will verify this afterwards.

cmake/llvm-toolchain.cmake
wusatosi: 👌 good suggestion

cmake/msvc-toolchain.cmake
wusatosi: Yes. 

Do I need to include documentation for this? 

Omitting this flag will cause warning C5072.

MSVC side documentation:
https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-c5072?view=msvc-170

cmake/llvm-toolchain.cmake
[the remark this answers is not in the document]
steve-downey: Cmake will process the toolchain file a random number of times if it isn't guarded. If you are careful and all the string operations are idempotent this is harmless, but it's a common source of bugs, as well as overhead in configuration.

cmake/gnu-toolchain.cmake
steve-downey: Tools like IDEs usually treat selecting the build type as a first class UI operation, where setting custom options is much more buried, so if running sanitizers is a build type, it's much easier to run them reliably that way. 

The alternative of using a distinct toolchain where e.g. memory sanitizer is part of the default build types makes it easier to apply the same options across many packages. That's particularly useful if you're using a package manager and getting pre-built libraries from them if they're available. Mixing thread or memory sanitizer libraries with ones without produces far too many false positives as the sanitizer can't do the tracking for memory initialization or lock acquisition/release in the libraries that aren't instrumented. 

Address and undefined behavior sanitizers don't have that tracing/tracking problem, so they're reasonably accurate if just the code under test is instrumented.

cmake/llvm-toolchain.cmake
steve-downey: This is my fault, to the extent this is coming from optional26. Address composes with many others, and I didn't try to find a better name for the collection. MaxSan is at least not wrong, and I don't have a better idea.

cmake/msvc-toolchain.cmake
steve-downey: If they fix it, we could remove the flag? "Why" comments are useful.

cmake/llvm-toolchain.cmake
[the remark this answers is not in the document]
wusatosi: @steve-downey I think I've seen what you've described in a previous iteration, where the toolchain gets executed multiple times adding duplicated flags. I don't think this happens on current toolchain script if we omit the include guard, but it seems unstable.

cmake/llvm-toolchain.cmake
wusatosi: Addressed.

cmake/llvm-toolchain.cmake
wusatosi: Adopted!

cmake/llvm-toolchain.cmake
wusatosi: Done

cmake/llvm-toolchain.cmake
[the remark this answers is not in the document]
wusatosi: Looks like this is not needed. I have omitted include guard for current toolchain file, we can include this in the future if needed.

wusatosi:
@camio this PR is ready for another round of review.

cmake/llvm-toolchain.cmake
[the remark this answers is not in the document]
steve-downey: That it doesn't produce incorrect results this time doesn't mean the template should drop it. It still adds overhead on reconfigure and can produce confusing bugs if it's not idempotent. 
Begin as you mean to proceed in the example we are providing as the basis for a project.

cmake/appleclang-toolchain.cmake
```diff
@@ -0,0 +1,37 @@
...
+        "-fsanitize=address -fsanitize=pointer-compare -fsanitize=pointer-subtract -fsanitize=undefined"
+    )
+elseif(BEMAN_BUILDSYS_SANITIZER STREQUAL "TSan")
+    set(SANITIZER_FLAGS "-fsanitize=thread")
+endif()
+
+set(CMAKE_C_FLAGS_DEBUG_INIT "${SANITIZER_FLAGS}")
+set(CMAKE_CXX_FLAGS_DEBUG_INIT "${SANITIZER_FLAGS}")
+
+set(RELEASE_FLAG "-O3 ${SANITIZER_FLAGS}")
```
camio, line 31: I think this variable should be prefixed with `BEMAN_BUILDSYS_` to avoid conflicts with project variables.

camio, line 31: FLAG -> FLAGS to match the existing CMake conventions.

camio (approved):
I left a couple minor naming suggestions, but otherwise this looks ready to ship.

cmake/appleclang-toolchain.cmake
wusatosi: This is only meant to be a local variable and not to be exported. I will see if I can make it so that it's treated like a local variable.

wusatosi: done!

wusatosi: I don't think this variable is exported out. Being a toolchain file, this shouldn't overwrite anything. `BEMAN_BUILDSYS_` implies this variable to be externally facing.

camio: Oops, you're right. I didn't realize that variables set in the toolchain don't propagate.

cmake/msvc-toolchain.cmake
wusatosi: Adopted

cmake/llvm-toolchain.cmake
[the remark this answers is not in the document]
wusatosi: @camio @steve-downey I am receiving conflicting feedback and I see both sides. What should I do?

steve-downey (approved):
This improves things to the point that we should land it. 

We can discuss what toolchain files should generally look like in follow ons.

cmake/llvm-toolchain.cmake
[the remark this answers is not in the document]
steve-downey: Not having the equivalent of include guards here isn't actually going to break anything with these toolchain files, we can discuss if and when the problem arises. 
As long as we avoid the cmake string concat idiom in what ends up being cached, and the operations are idempotent, processing the toolchain file more than strictly necessary is mostly harmless.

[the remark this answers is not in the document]
camio: I'm surprised the toolchain gets executed multiple times. That's too bad. I agree with the argument that the include guard should be kept in that case.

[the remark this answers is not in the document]
wusatosi: This is added back.

PR #77, opened 2024-11-14
Remove `include(ctest)`

### LINKED ISSUE #60: Do only include(CTest) if dashboard is needed
See https://cmake.org/cmake/help/latest/manual/ctest.1.html

wusatosi:
Fixes #60 .

I am not really understanding if this is desired, so reject this if this change is not wanted.

Explicitly asking @camio @bretbrownjr and @ClausKlein for input.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,7 +27,6 @@ include(FetchContent)
 include(GNUInstallDirs)
 
 if(BEMAN_EXEMPLAR_BUILD_TESTING)
-
     enable_testing()
 
     # Fetch GoogleTest
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,7 +26,6 @@ option(
 include(GNUInstallDirs)
 
 if(BEMAN_EXEMPLAR_BUILD_TESTING)
-    include(CTest)
     include(FetchContent)
 
     enable_testing()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,11 +23,12 @@ option(
     ${PROJECT_IS_TOP_LEVEL}
 )
 
-include(CTest)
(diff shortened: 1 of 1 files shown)


steve-downey (approved):
👍

camio (commented):
I think we can drop `include(CTest)` entirely. Also, the code was simpler when FetchContent wasn't conditionally included.

wusatosi:
> [we can drop `include(CTest)` entirely]

Done in CMakeLists, the CI isn't failing so I guess this would work.

Will this impact commands such as `ctest --test-dir build --build-config Release`?

CMakeLists.txt
```diff
@@ -23,11 +23,11 @@ option(
     ${PROJECT_IS_TOP_LEVEL}
 )
 
-include(CTest)
-include(FetchContent)
 include(GNUInstallDirs)
 
 if(BEMAN_EXEMPLAR_BUILD_TESTING)
+    include(FetchContent)
```
camio, line 29: Sorry for not elaborating on my rationale before. Keeping all includes near the top of the file makes it easy to determine a file's dependencies because they're all clustered together. We have a similar convention with C/C++ includes.

```suggestion
include(FetchContent)
include(GNUInstallDirs)

if(BEMAN_EXEMPLAR_BUILD_TESTING)
```

I approve the PR, but would like this change applied.

camio (approved):

camio:
>>  [we can drop include(CTest) entirely]
>
> Will this impact commands such as ctest --test-dir build --build-config Release?

Nope, we no longer need anything this module provides.

CMakeLists.txt
wusatosi: Thanks for the context!

PR #66, opened 2024-11-01
Feature/support macos builds

ClausKlein:
Edit from river:
Closes #22 .
Closes #64 .
Closes #65 .
Closes #67 .

PR #80, opened 2024-11-19
Introduce MSVC CI support

### LINKED ISSUE #24: MSVC CI Support
It's important for Beman projects to support MSVC in continuous integration, especially since not all contributors will have immediate access to that platform.

This issue will be done when the main branch of this repo includes at least one MSVC flavored build.

wusatosi:
This PR adds MSVC CI support.

Could be improved after #76 is merged.

Conflict with unknown state PR #66 .

Closes #24 .

### DIFF
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -104,10 +104,18 @@ jobs:
         with:
           arch: x64
       - name: Print installed softwares
+        shell: bash
         run: |
           echo "Compiler:"
-          ${{ matrix.platform.cpp }} --version
-          ${{ matrix.platform.c }} --version
+
+          # cl does not have a --version option
+          if [ "${{ matrix.platform.cpp }}" != "cl" ]; then
+            ${{ matrix.platform.cpp }} --version
+            ${{ matrix.platform.c }} --version
+          else
+            ${{ matrix.platform.cpp }}
+            ${{ matrix.platform.c }}
+          fi
 
           echo "Build system:"
           cmake --version
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -174,14 +174,15 @@ jobs:
         run: |
           cmake --build build --config Release --verbose
           cmake --build build --config Release --target all_verify_interface_header_sets
-          cmake --install build --config Release --prefix /opt/beman.exemplar
-          find /opt/beman.exemplar -type f
+          cmake --install build --config Release --prefix ${{ runner.temp }}/beman.exemplar
+          ls -R ${{ runner.temp }}/beman.exemplar
       - name: Build Debug
         run: |
           cmake --build build --config Debug --verbose
           cmake --build build --config Debug --target all_verify_interface_header_sets
           cmake --install build --config Debug --prefix /opt/beman.exemplar
-          find /opt/beman.exemplar -type f
+          cmake --install build --config Release --prefix ${{ runner.temp }}/beman.exemplar
+          ls -R ${{ runner.temp }}/beman.exemplar
 
   create-issue-when-fault:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -123,18 +123,16 @@ jobs:
         run: |
           cmake --build build --config Release --verbose
           cmake --build build --config Release --target all_verify_interface_header_sets
-          cmake --install build --config Release --prefix /opt/beman.exemplar
-          cd /opt/beman.exemplar
-          tree
-          cd -
+          cmake --install build --config Release --prefix ${{ runner.temp }}/beman.exemplar
+          tree ${{ runner.temp }}/beman.exemplar
       - name: Test Release
         run: ctest --test-dir build --build-config Release
       - name: Build Debug
         run: |
           cmake --build build --config Debug --verbose
           cmake --build build --config Debug --target all_verify_interface_header_sets
-          cmake --install build --config Debug --prefix /opt/beman.exemplar
-          tree /opt/beman.exemplar
+          cmake --install build --config Debug --prefix ${{ runner.temp }}/beman.exemplar
+          tree ${{ runner.temp }}/beman.exemplar
       - name: Test Debug
         run: ctest --test-dir build --build-config Debug
 
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -126,7 +126,9 @@ jobs:
           cmake --build build --config Release --verbose
           cmake --build build --config Release --target all_verify_interface_header_sets
           cmake --install build --config Release --prefix /opt/beman.exemplar
-          tree /opt/beman.exemplar
+          cd /opt/beman.exemplar
+          tree
+          cd -
       - name: Test Release
         run: ctest --test-dir build --build-config Release
       - name: Build Debug
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -124,7 +124,7 @@ jobs:
           cmake --build build --config Release --verbose
           cmake --build build --config Release --target all_verify_interface_header_sets
           cmake --install build --config Release --prefix ${{ runner.temp }}/beman.exemplar
-          tree ${{ runner.temp }}/beman.exemplar
+          ls -R ${{ runner.temp }}/beman.exemplar
       - name: Test Release
         run: ctest --test-dir build --build-config Release
       - name: Build Debug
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_tests.yml
```diff
@@ -46,59 +49,92 @@ jobs:
...
+              description: "ubuntu gcc"
               cpp: g++
               c: gcc
+              os: ubuntu-latest
             cpp_version: 17
             cmake_args:
               description: "Dynamic"
               args: "-DBUILD_SHARED_LIBS=on"
+            # MSVC has limited support for sanitizers, so we use include here for MSVC as an exception.
+            # This should be simplified after beman-project/exemplar#76 is merged.
```
camio, line 71:
```suggestion
```

It isn't clear what is meant by "use include here". Also, it is unclear if #76 in its current form will be merged.

.github/workflows/ci_tests.yml
```diff
@@ -46,59 +49,92 @@ jobs:
...
+          CC: ${{ matrix.platform.c }}
+          CXX: ${{ matrix.platform.cpp }}
           CMAKE_GENERATOR: "Ninja Multi-Config"
       - name: Build Release
         run: |
           cmake --build build --config Release --verbose
           cmake --build build --config Release --target all_verify_interface_header_sets
           cmake --install build --config Release --prefix /opt/beman.exemplar
-          find /opt/beman.exemplar -type f
+          tree /opt/beman.exemplar
```
camio, line 129: Maybe something like this?
```suggestion
          cd /opt/beman.exemplar
          tree
          cd -
```

.github/workflows/ci_tests.yml
```diff
@@ -46,59 +49,92 @@ jobs:
...
+        uses: TheMrMilchmann/setup-msvc-dev@v3
+        with:
+          arch: x64
       - name: Print installed softwares
         run: |
-          clang++ --version
-          g++ --version
+          echo "Compiler:"
+          ${{ matrix.platform.cpp }} --version
+          ${{ matrix.platform.c }} --version
```
camio, line 112: Note that `--version` isn't a `cl.exe` flag. Maybe something like this? See [this reference page](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#operators).

```suggestion
          ${{ matrix.platform.cpp }} ${{ matrix.platform.cpp == 'cl' && '' || '--version' }}
          ${{ matrix.platform.c }}  ${{ matrix.platform.cpp == 'cl' && '' || '--version' }}
```

camio (commented):
A couple nits on platform-independent syntax, but otherwise looks good.

.github/workflows/ci_tests.yml
wusatosi: that's a great idea!

.github/workflows/ci_tests.yml
wusatosi: I don't think this is working.

Also `--version` seem to work? (I know this shouldn't but it's working).
image

.github/workflows/ci_tests.yml
wusatosi: yeah sorry I should check my comment super well.

.github/workflows/ci_tests.yml
wusatosi: Switched to a GitHub action runner provided directory. Still got empty directories from tree command, I don't know if `tree` is just badly implemented under bash environment

wusatosi: Agh, it's a quirk with powershell tree.

See: https://superuser.com/questions/1270040/windows-10-powershell-tree-command-how-to-show-files

wusatosi: switch out to `ls -R`, and it is working. It's just very ugly...

image
image

.github/workflows/ci_tests.yml
wusatosi: done.

.github/workflows/ci_tests.yml
```diff
@@ -46,59 +49,98 @@ jobs:
...
+          CC: ${{ matrix.platform.c }}
+          CXX: ${{ matrix.platform.cpp }}
           CMAKE_GENERATOR: "Ninja Multi-Config"
       - name: Build Release
         run: |
           cmake --build build --config Release --verbose
           cmake --build build --config Release --target all_verify_interface_header_sets
           cmake --install build --config Release --prefix /opt/beman.exemplar
-          find /opt/beman.exemplar -type f
+          ls -R /opt/beman.exemplar
```
neatudarius, line 135: why `find` is not good?

.github/workflows/ci_tests.yml
```diff
@@ -31,12 +31,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        platform: [ubuntu-latest]
-        compiler:
-          - cpp: g++
+        platform:
+          - description: "ubuntu gcc"
```
neatudarius, line 35:
```suggestion
          - description: "Ubuntu GCC"
```
if it's in description?

.github/workflows/ci_tests.yml
```diff
@@ -31,12 +31,15 @@ jobs:
...
+            cpp: g++
             c: gcc
```
neatudarius, line 37: Do we ever use C compiler? More a curiosity.

.github/workflows/ci_tests.yml
wusatosi: Because we need to execute this on windows runner for MSVC, where tree doesn't work with the /opt/.... Directory.

Should I include a comment here documenting this?

neatudarius: OK, got it, right.  I would add a command like `name: Build Release # Portable commands only`. thanks!

.github/workflows/ci_tests.yml
wusatosi: This is carried over from previous commits. 

We can evaluate this later

steve-downey: > Do we ever use C compiler? More a curiosity.

Probably not us deliberately, but Cmake will look for one anyway, and any packages we fetch might use it. When this goes wrong it's surprisingly difficult to track down.

neatudarius: Let's keep it.

.github/workflows/ci_tests.yml
camio: You can use the same trick I suggested in [this comment](https://github.com/bemanproject/exemplar/pull/80#discussion_r1870168036) to use the `/F` flag on windows only

.github/workflows/ci_tests.yml
wusatosi: Adopted.

.github/workflows/ci_tests.yml
wusatosi: Adopted.

camio (approved):
The `# Portable commands only` comments clutter up the code in my opinion, but I don't want to hold up merging for that :-D

wusatosi:
> The `# Portable commands only` comments clutter up the code in my opinion, but I don't want to hold up merging for that :-D

Yeah I do agree it clutters up the code...

.github/workflows/ci_tests.yml
```diff
@@ -46,59 +49,100 @@ jobs:
...
-          find /opt/beman.exemplar -type f
+          ls -R /opt/beman.exemplar
       - name: Test Release
         run: ctest --test-dir build --build-config Release
       - name: Build Debug
         run: |
+          # Portable commands only
           cmake --build build --config Debug --verbose
           cmake --build build --config Debug --target all_verify_interface_header_sets
           cmake --install build --config Debug --prefix /opt/beman.exemplar
```
steve-downey, line 144: This install isn't going to do anything because it's installing to the same location as the Release install. Unless that's the intent?

wusatosi: Intention here is to test the install command.

steve-downey: It ends up skipping installation because it's already installed, which is a sort of test. Not critical for header only, but if there's any generated config or binaries, they shouldn't clobber each other.

steve-downey:
Perhaps an explanatory general comment, once, reminding us to use the portable subset of commands. 
On the other hand, if it breaks in CI, a PR that doesn't use the portable subset will fail?

wusatosi:
> Perhaps an explanatory general comment, once, reminding us to use the portable subset of commands. On the other hand, if it breaks in CI, a PR that doesn't use the portable subset will fail?

Ops I didn't see this before merging. I don't think this is fine. We can move this out in a later PR.

Issue #73, opened 2024-11-11
Enable latest versions of compilers on main


neatudarius:
Enable latest versions of compilers on main

- [ ] Enable gcc-15
- [ ] Enable gcc-14
- [ ] Enable gcc-13
- [ ] Enable gcc-12
- [ ] Enable clang-20
- [ ] Enable clang-19
- [ ] Enable clang-18
- [ ] Enable clang-17
- [ ] Enable clang-16
- [ ] TBD msvc

wusatosi:
Are you planning to shove all these into the CI?

We can create a nightly CI job for this, what will the build parameters be?
Note that building this with current version matrix (17,20,23,26)  will result in a 40 job matrix.

neatudarius:
Well, I think we need to discuss.

We have an action from our meetings (1-2 months ago) to define a matrix for environments.

It's absolutely cool that we have existing matrix, which I understand it just run "default " version of gcc/clang, but we also need to make sure we have support on a wide range of compiler version, in my opinion.

Example latest 3-4 compiler version for gcc clang, msvc.

wusatosi:
You can actually see the time spent for actions in this tab: [link](https://github.com/beman-project/exemplar/actions/metrics/performance).

Currently each job (one combination of the matrix, basically per compile command variant per C++ version per compiler) takes 29s to run, setup time is usually ~5s. 
With 28 jobs, it takes GitHub Actions ~1 min - ~2 mins to run them, we are limited by queuing time (GitHub load) + parallelization.

I do want to point out that we do not have a lot to test in this test case, 29s could potentially be higher if the library consumer is developing complex library (longer compile time)/ library with timeout (longer testing time). Also we are going to test more platforms (macos, windows), this would dramatically increase the number of combinations.

I think we are on free tier of GitHub so our max parallelization is 20 jobs, see: [here](https://docs.github.com/en/actions/administering-github-actions/usage-limits-billing-and-administration#usage-limits).
I believe we are not paying GitHub, open source project gets free tire.

My point is, I don't think it will be reasonable to trigger another matrix of (per compiler version per c++ version potentially per compile option variant per platform) on each commit, maybe only trigger (on main branch + in pr) / nightly.

Though realistically, it will probably take ~5 mins to run our current matrix + macos matrix + compiler matrix (assume one per version per compiler), this is definitely not horrible but not desirable as well.

ClausKlein:
[...]

PR #84, opened 2024-11-26
Compiler test

wusatosi:
Implement trivial cases for #73 .

By trivial I mean either available from package managers (GCC), or via officially supported install script (LLVM).

Included versions:

GCC:
- 14
- 13
- 12
Note: Lower versions are generally available on apt. GCC 15 mentioned in #73 is not available due to no official release. 

LLVM:
- 20
- 19
- 18
- 17
Note: Lower versions (#73 requests 16) is not available from llvm's official script as it is no longer supported. See: https://github.com/llvm/llvm-project/issues/90536

This adds 7 new CI jobs.

As GCC/ Clang doesn't publish their nightly build on package managers, using nightly/ head of compiler will need us to build either a custom image or be tested in a nightly job.

Note: A more complicated but CI-wide salable approach using custom build docker image is being experimented.
Instead of pulling various versions of compiler on every CI run, we use a pre-built docker image that has all compilers install.
See: https://github.com/wusatosi/cpp-docker/ for an example of docker image, we can speed this up by having multiple kinds of image.
See: https://github.com/wusatosi/exemplar/tree/compiler-test-docker as a proof of concept on exemplar side implementation.
See: https://github.com/wusatosi/exemplar/actions/runs/12216494307/job/34079670273 as an example of actions run.

### DIFF
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -182,9 +182,6 @@ jobs:
 
             $CC --version
             $CXX --version
-
-            echo "CC=$CC" >> "$GITHUB_OUTPUT"
-            echo "CXX=$CXX" >> "$GITHUB_OUTPUT"
           else
             wget https://apt.llvm.org/llvm.sh
             chmod +x llvm.sh
@@ -195,10 +192,10 @@ jobs:
 
             $CC --version
             $CXX --version
-
-            echo "CC=$CC" >> "$GITHUB_OUTPUT"
-            echo "CXX=$CXX" >> "$GITHUB_OUTPUT"
           fi
+
+          echo "CC=$CC" >> "$GITHUB_OUTPUT"
+          echo "CXX=$CXX" >> "$GITHUB_OUTPUT"
       - name: Configure CMake
         run: |
           cmake -B build -S . -DCMAKE_CXX_STANDARD=20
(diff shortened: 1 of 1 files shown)


steve-downey:
Don't do this to your own system, it will eventually corrupt your OS. However, for a throwaway VM like CI:
```
sudo add-apt-repository ppa:ubuntu-toolchain-r/test
sudo apt-get update
sudo apt-get install gcc-15 g++-15
```
or something similar 

The ubuntu-toolchain-r/test apt repo is the Ubuntu toolchain group's test repository.

wusatosi:
Just out of curiosity, why would it break your os?

steve-downey:
Not every time, and not immediately, but it replaces components with the test versions, and eventually gets something slightly wrong in such a way that apt can't generate a solution for normal updates and installs. You can generally recover by uninstalling everything and reinstalling without the PPA.

wusatosi:
> Don't do this to your own system, it will eventually corrupt your OS. However, for a throwaway VM like CI:
> 
> ```
> sudo add-apt-repository ppa:ubuntu-toolchain-r/test
> sudo apt-get update
> sudo apt-get install gcc-15 g++-15
> ```
> 
> or something similar
> 
> The ubuntu-toolchain-r/test apt repo is the Ubuntu toolchain group's test repository.

Hum, I don't think its working.
Reports: `E: Unable to locate package gcc-15`

Maybe gcc-15 is not available on 24.04?

See: https://github.com/bemanproject/exemplar/actions/runs/12215966614/job/34078564865

wusatosi:
@neatudarius this is requested by you, do you mind reviewing this PR?

neatudarius (approved):

PR #88, opened 2024-12-05
remove unnecessary comments

camio:
These comments add clutter in my opinion. If someone is editing this file,
they can search google to learn more information about pre-commit. We
don't, for example, link to the CMake homepage from our CMakeLists.txt.

Additionally, the other comments here are pieces of reference documentation
elsewhere.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,3 @@
-# See https://pre-commit.com for more information
-# See https://pre-commit.com/hooks.html for more hooks
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v5.0.0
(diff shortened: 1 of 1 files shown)


wusatosi:
Why are these comments considered unnecessary?

camio:
> Why are these comments considered unnecessary?

Please see updated PR text.

.pre-commit-config.yaml
```diff
@@ -8,25 +6,16 @@ repos:
...
-
-    # CMake linting and formatting
   - repo: https://github.com/BlankSpruce/gersemi
     rev: 0.15.1
     hooks:
     - id: gersemi
       name: CMake linting
-
-    # Markdown linting
-    # Config file: .markdownlint.yaml
```
wusatosi, line 29: There are multiple CMake linters using different file names for their linting file. This is useful for alleviate confusion when someone brings in another CMake lint config.

wusatosi:
> > Why are these comments considered unnecessary?
> 
> Please see updated PR text.

See review comment.

neatudarius (approved):

.pre-commit-config.yaml
wusatosi: You can see an instance of this happening here:
https://github.com/bemanproject/inplace_vector/pull/14/files#r1842893309

.pre-commit-config.yaml
```diff
@@ -8,25 +6,16 @@ repos:
     -   id: end-of-file-fixer
     -   id: check-yaml
     -   id: check-added-large-files
-
-    # Clang-format for C++
-    # This brings in a portable version of clang-format.
```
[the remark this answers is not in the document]
wusatosi: I mean they can certainly search for it. But I don't think this is non-dependency is intuitive.

camio:
Alright, I'm withdrawing this PR for now since there lacks consensus for it.

I maintain that the file is perfectly understandable without the comments since a) one can follow the link right below the comment to understand the semantics of what is being brought in, and b) googling for .pre-commit-config.yaml provides all the links necessary for understanding the utility of this file.

> I also think exemplar should be as descriptive as it can be. We create exemplar so people don't have to search stuff up.

Exemplar's purpose is to exemplify best practices applied to a C++ repository, not to be a descriptive tutorial. Just as it is an anti-pattern to redundantly document the semantics of functions at their call sites, we should not document our tools where they are being used.

PR #92, opened 2024-12-28
Support system libs

bretbrownjr:
Problem
-------

**Given**:

A sandbox environment is an environment where dependencies
are all provided but access to the internet is not. Also take as
a given that the build for this project happens in an environment
that meets that description.
    
**When**:
    
Configuring this project with any of the provided approaches. For
instance:
    
```
cmake -B build -S .
```
    
**Then**:
    
The configuration step hangs or fails, depending on the nature of
the sandboxed environment.
    
Solution
--------
    
Alter `FetchContent_*` calls in this project to use `GTest` as the
project name and nothing else. This is important to allow the
`FETCHCONENT_TRY_FIND_PACKAGE_MODE` feature to work as designed.
This allows the human or tool building this project to explicitly
disable operations involving calls to `git` or other off-box I/O.
    
The documentation in the README was also updated to include
workflow tweaks for building against packaged versions of either
existing GoogleTest source code or prebuilt GoogleTest libraries.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,14 +31,14 @@ include(GNUInstallDirs)
 if(BEMAN_EXEMPLAR_BUILD_TESTS)
     # Fetch GoogleTest
     FetchContent_Declare(
-        googletest
+        GTest
         GIT_REPOSITORY https://github.com/google/googletest.git
         GIT_TAG
             f8d7d77c06936315286eb55f8de22cd23c188571 # release-1.14.0
         EXCLUDE_FROM_ALL
     )
     set(INSTALL_GTEST OFF) # Disable GoogleTest installation
-    FetchContent_MakeAvailable(googletest)
+    FetchContent_MakeAvailable(GTest)
 endif()
 
 add_subdirectory(src/beman/exemplar)
--- a/README.md
+++ b/README.md
@@ -325,6 +325,30 @@ cmake -B build -S . -DBEMAN_EXEMPLAR_BUILD_TESTING=OFF
 
 </details>
 
+<details>
+<summary> Disable `git clone` operations </summary>
+
+By default, while tests are enabled, this project will clone GoogleTest
+and build it from source. In many environments, it is better to use an
+already available GoogleTest instead.
+
+### Existing Source Code
+
+For instance, in Ubuntu, source code for GoogleTest is provided via a
+debian-style package under the directory `/usr/src/googletest`. To
+use those sources instead of cloning other sources from the internet,
+provide `-DFETCHCONTENT_SOURCE_DIR_GTEST=/usr/src/googletest` as a
+configuration argument when calling `cmake`.
+
+### Existing Binaries
+
+In another case, again in Ubuntu, GoogleTest can be installed prebuilt
+with CMake discovery support deployed in normal search paths.
+Setting `-DFETCHCONTENT_TRY_FIND_PACKAGE_MODE=ALWAYS` when calling `cmake`
+will force the build of this project to locate that installed library.
+
+</details>
+
 ## Integrate beman.exemplar into your project
 
 <details>
@@ -378,6 +402,7 @@ Build systems that support `pkg-config` by providing a `beman.exemplar.pc` file.
 Build systems that support interoperation via `pkg-config` should be able to detect `beman.exemplar` for you automatically.
 
 </details>
+
 
 ## Contributing
 


bretbrownjr:
This is ready to review, though it includes the change in #91, so I'll leave it in draft for now. We could either merge this PR with both changes and close #91 or rebase this change on top of #91 after it is merged.

CMakeLists.txt
```diff
@@ -9,38 +9,46 @@ project(
...
+            GIT_TAG
+                f8d7d77c06936315286eb55f8de22cd23c188571 # release-1.14.0
+            EXCLUDE_FROM_ALL
+        )
+        block()
+            set(INSTALL_GTEST OFF) # Disable GoogleTest installation
+            FetchContent_MakeAvailable(googletest)
+        endblock()
+    else()
+        find_package(GTest REQUIRED)
```
JeffGarland, line 50: Is there never a case where we don't need google test?  I'm actually hoping projects use something else mostly

JeffGarland:
except for my question, I have no issues with the PR

CMakeLists.txt
bretbrownjr: Projects that test another way wouldn't need googletest, I suppose. In particular, projects providing entirely constexpr features wouldn't benefit from a framework like googletest, so the extra dependency would get removed. Also, some libraries would be too low level to justify a dependency on a test framework that pulls in vectors and serializers and such.

I would still expect a CMake test target to exercise a "test suite" of some sort, even if that's a series of example main functions that get compiled and executed.

wusatosi, line 50: Would this cause version conflict? e.g. we are fetching a specific version of Google Test, but find package finds a different (e.g. lower version) of it.

JeffGarland: > entirely constexpr features wouldn't benefit from a framework 

Well, in my experience with this in boost date-time I still had runtime tests along with the constexpr validation tests -- of course in that case the runtime tests came first and constexpr later.  Still, with a trust-but-verify thinking I'd still want both compile and runtime covered -- but I wouldn't need/use the massive google test for simple test cases.

At the moment I'm not sure how to resolve the issue -- perhaps we leave a comment that says 'replace test framework here' -- not sure.

camio:
FetchContent already has this functionality built-in. I don't think we need this PR. See [FETCHCONTENT_TRY_FIND_PACKAGE_MODE](https://cmake.org/cmake/help/latest/module/FetchContent.html#variable:FETCHCONTENT_TRY_FIND_PACKAGE_MODE).

JeffGarland:
Interesting -- would it be more correct to say we need a *different PR*? I think there's still a bit of an issue in the original cmake code.

camio:
> I think there's still a bit of an issue in the original cmake code.

Can you elaborate on that?

JeffGarland:
yeah, sorry -- what I mean is the file as is unconditionally requires google test -- so one might get the impression that it's a requirement for any testing when it's not.  I suppose it's fine to leave for now. @camio are you suggesting we abandon this PR?

camio:
> what I mean is the file as is unconditionally requires google test

The file as-is only pulls in google test when `BEMAN_EXEMPLAR_BUILD_TESTS` is set to `ON` and this cmake option defaults to ON only when the project is top-level.

> @camio are you suggesting we abandon this PR?

Yes

CMakeLists.txt
bretbrownjr: I think you have valid concerns, Jeff. Let's keep iterating until we have something we're happy with.

That being said, the current PR is entirely about not getting the build to hang given the dependency. Are OK branching off another discussion so we can resolve the bug noted in the meantime?

bretbrownjr:
I don't believe the failing `msvc-debug` version of this build should have been affected by this change. @wusatosi or @neatudarius can you confirm?

wusatosi:
> I don't believe the failing `msvc-debug` version of this build should have been affected by this change. @ wusatosi or @ neatudarius can you confirm?

Looks like the MSVC family of compiler is having trouble finding google test dependency. 

> D:\a\exemplar\exemplar\build\msvc-debug\_deps\gtest-src\googletest\src\gtest-all.cc(38): fatal error C1083: Cannot open include file: 'gtest/gtest.h': No such file or directory

I will try to investigate.

_btw bret can you check your email?_

README.md
```diff
@@ -325,6 +325,30 @@ cmake -B build -S . -DBEMAN_EXEMPLAR_BUILD_TESTING=OFF
 
 </details>
 
+<details>
+<summary> Disable `git clone` operations </summary>
```
wusatosi, line 329: Nit: maybe `Disable pulling Google Test from GitHub` would be a more clear summary here.

README.md
```diff
@@ -325,6 +325,30 @@ cmake -B build -S . -DBEMAN_EXEMPLAR_BUILD_TESTING=OFF
...
+By default, while tests are enabled, this project will clone GoogleTest
+and build it from source. In many environments, it is better to use an
+already available GoogleTest instead.
+
+### Existing Source Code
+
+For instance, in Ubuntu, source code for GoogleTest is provided via a
+debian-style package under the directory `/usr/src/googletest`. To
+use those sources instead of cloning other sources from the internet,
+provide `-DFETCHCONTENT_SOURCE_DIR_GTEST=/usr/src/googletest` as a
```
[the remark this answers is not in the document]
wusatosi: You could technically config a void proxy for git and then test it.

See: https://gist.github.com/evantoli/f8c23a37eb3558ab8765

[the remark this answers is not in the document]
bretbrownjr: That would involve interesting CI changes in each case, possibly on another PR?

- The Ubuntu images would need the GoogleTest package installed for these flags to work as designed
- We haven't selected a package manager on Windows -- though we should add support for building against Conan and vcpkg dependencies (GoogleTest in this case) as you and I discussed in person a little bit yesterday

Given we're dealing with an existing bug, I think it's OK to resolve the bug (once we can figure out why Windows is being weird) and circle back on a regression prevention test, though maybe @bemanproject/leads want to chime in.

README.md
bretbrownjr: It's not specifically GitHub that is the issue -- it's the off-box I/O as such.

Is the confusion that it's not obvious that `git clone` is being invoked by `FetchContent`? Possibly we need to update some comments in the `CMakeLists.txt` to make that more clear.

README.md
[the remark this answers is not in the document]
wusatosi: That's fair, I thought it would be as easy as `git clone https:github.com/google/googletest /usr/libs/googletest`

I usually do prefer having all CMake parameters tested (I don't think multi platform is that important here, the goal is to show the project could build under the parameters). When they break, they break silently, which is very annoying.

steve-downey:
Ubuntu currently will install googletest into /usr/src if you apt install
it. It should be locateable once that's done. We do a bunch of package
install already, one more should not be a problem.

README.md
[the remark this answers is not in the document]
bretbrownjr: > That's fair, I thought it would be as easy as git clone https:github.com/google/googletest /usr/libs/googletest

Fair. That's technically an exercise of the workflow for one of the two documented use cases. Though it's the functional equivalent of re-implementing a worse version of `FetchContent`, so I'd still rather hold off until we have packaged GoogleTest source to build against as such. If for no other reason, to leave time to work on more meaningful problems elsewhere in this repo.

bretbrownjr:
Just to save people time combing through the logs. Here is a more technical and detailed bug description for the CI failing here.

The failing compile command is:

```
C:\PROGRA~1\MICROS~2\2022\ENTERP~1\VC\Tools\MSVC\1442~1.344\bin\Hostx64\x64\cl.exe  /nologo /TP  -ID:\a\exemplar\exemplar\build\msvc-debug\_deps\gtest-src\googlemock\include -ID:\a\exemplar\exemplar\build\msvc-debug\_deps\gtest-src\googlemock -ID:\a\exemplar\exemplar\build\msvc-debug\_deps\gtest-src\include -ID:\a\exemplar\exemplar\build\msvc-debug\_deps\gtest-src /DWIN32 /D_WINDOWS   /permissive- /fsanitize=address /Zi /Ob0 /Od /RTC1 -MDd -Zi -GS -W4 -WX -wd4251 -wd4275 -nologo -J -D_UNICODE -DUNICODE -DWIN32 -D_WIN32 -DSTRICT -DWIN32_LEAN_AND_MEAN -wd4702 -utf-8 -DGTEST_HAS_PTHREAD=0 -EHsc -D_HAS_EXCEPTIONS=1  -std:c++20 /showIncludes /Fo_deps\gtest-build\googlemock\CMakeFiles\gmock_main.dir\src\gmock-all.cc.obj /Fdlib\gmock_mainpdb_debug_postfix-NOTFOUND.pdb /FS -c D:\a\exemplar\exemplar\build\msvc-debug\_deps\gtest-src\googlemock\src\gmock-all.cc
```

The error in the log is:

```
D:\a\exemplar\exemplar\build\msvc-debug\_deps\gtest-src\googlemock\include\gmock/internal/gmock-port.h(57): fatal error C1083: Cannot open include file: 'gtest/internal/gtest-port.h': No such file or directory
```

Notably the source file being built is at:

```
D:\a\exemplar\exemplar\build\msvc-debug\_deps\gtest-src\googlemock\src\gmock-all.cc
```

Which would lead me to believe we would expect to see an include directory like this in the compile command:

```
-ID:\a\exemplar\exemplar\build\msvc-debug\_deps\gtest-src\googletest\include
```

See [the relevant path in the googletest repo](https://github.com/google/googletest/tree/main/googletest/include) for reference.

But we don't see that include path. We see these instead, presumably intended to point at the `googletest` headers:

```
-ID:\a\exemplar\exemplar\build\msvc-debug\_deps\gtest-src\include
-ID:\a\exemplar\exemplar\build\msvc-debug\_deps\gtest-src
```

bretbrownjr:
So. @camio. I think using this approach to support `FetchContent` and `find_package` is actually broken for arcane CMake reasons involving name collisions. I'll try to explain.

First, the latest commit changes the `FetchContent` "name" from `googletest` to `GTest`. This is required because that name must match the `find_package` "PackageName" for `FetchContent` to be able to call a `find_package` that actually resolve to the `FindGTest.cmake` file it expects to find.

Second, `googletest` use a variable named `gtest_SOURCE_DIR` to set its include paths using `include_directories(...)` [link](https://github.com/google/googletest/blob/main/googletest/CMakeLists.txt#L110-L112) (aside: Boo! Bad CMake practice compared to  `target_include_directories`).

Third, `FetchConent` will define `<name>_SOURCE_DIR` as part of its behavior [[doc link](https://cmake.org/cmake/help/latest/module/FetchContent.html#command:fetchcontent_makeavailable)].

Fourth, the `project(gtest ...)` call in `googletest` [[source link](https://github.com/google/googletest/blob/main/googletest/CMakeLists.txt#L50)] will define `gtest_SOURCE_DIR` since `gtest` is the project name [[doc link](https://cmake.org/cmake/help/latest/variable/PROJECT-NAME_SOURCE_DIR.html)].

Fifth, it's worth noting that `googletest` is actually two standalone CMake projects colocated in one git repo that is a superproject stitching them together. So `FetchContent` sets `gtest_SOURCE_DIR` to point to the directory *above* where the `googletest/` source code actually resides.

Finally, CMake is case insensitive largely, so don't get hung up on the distinction between "GTest" and "gtest" in correlating the above.

Anyway, all this leads me to believe that, in the currently proposed patch:

* `FetchContent` is declaring `gtest_SOURCE_DIR` to point to the top-level directory of the `github:googletest/googletest`

* Eventually the `googletest/` subdir (a.k.a. `gtest` CMake "project") runs, using that value of `gtest_SOURCE_DIR` instead of the actual project directory for that project.

* The include directories are therefore missing the `googletest/` subdir and we get missing includes.

This concludes my submission for the 2025 international CMake debugging championships.

bretbrownjr:
As to solutions... my personal opinion is and has been that putting `FetchContent` calls in regular `CMakeLists.txt` is a code smell and this debugging session hasn't softened that opinion.

To that end, we could move logic that provides `GTest::Main`, etc., to an `include(resolve_dependencies.cmake)` or somesuch that allows us to more gracefully and semantically select between vendoring github source code into the build versus discovering prebuilt GoogleTest on the system. Either immediately or eventually we could put hardcoded dependency info (URLs, names, etc.) in a JSON file and not in CMake.

The option that is the least amount of work is to go back to the previous implementation on this PR -- reintroduce a user-facing option to use FetchContent or not for GoogleTest.

bretbrownjr:
Oh, and as to why only failing in Windows... case insensitive filesystems?

bretbrownjr:
Since there are design and/or CMake practices questions now, this PR is not ready to merge. I'm moving it back into draft status.

camio:
Apparently, this is a known defect in Google test and they're open to PRs that will fix it (See https://github.com/google/googletest/issues/4384). It doesn't seem that hard to essentially rename that variable.

camio:
I changed the FetchContent name to GTest and bumped the GTest version in #128. That worked.

camio:
I also [started a discussion](https://discourse.cmake.org/t/fetchcontents-googletest-example-causes-confusion/13651) on the CMake forums about this.

neatudarius (approved):

bretbrownjr:
Addressed by #105

PR #93, opened 2024-12-31
Add new_project_from_exemplar.py.

tzlaine:
As discussed in this thread on Discourse: https://discourse.bemanproject.org/t/working-on-new-project-from-exemplar-py/293 .

### DIFF
--- a/new_project_from_exemplar.py
+++ b/new_project_from_exemplar.py
@@ -1,0 +1,288 @@
+#!/usr/bin/env python3
+
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+import argparse
+import getpass
+import os
+import sys
+try:
+    from git import Repo
+except:
+    print("You do not have GitPython installed, but it is required.  Ubuntu users can install with 'sudo apt-get install python3-git'.  Other users, see https://gitpython.readthedocs.io/en/stable/intro.html.")
+    sys.exit(1)
+
+parser = argparse.ArgumentParser(description="Adapts the Beman exemplar project for your use as a brand new project.  Leaves 'TODO' in several places that need you attention (especially in README.md).")
+parser.add_argument('project_name', type=str, help="The name of your new project.  This should be the same name as the Github repo name.  For example, 'my_proj' in 'git@github.com:me/my_proj.git'")
+parser.add_argument('--owner', type=str, default=getpass.getuser(), help="The user or group in which this project's Github repo is found.  For example, 'me' in 'git@github.com:me/my_proj.git'.  Defaults to your username on this system.")
+parser.add_argument('--paper', type=str, default="TODO", help="The paper that this Beman project is implementing (e.g. P1234R5).")
+parser.add_argument('--cpp-version', type=int, default=26, help="The C++ version required to build this project.")
+parser.add_argument('--desc', type=str, default="TODO", help="The description of this project that should appear in the title of your README.md.")
+args = parser.parse_args()
+
+subtrees = ['src/beman', 'include/beman', 'tests/beman']
+
+# Before making any concrete changes, move 'exemplar' subtrees via 'git mv'.
+repo = Repo('.')
+for tree in subtrees:
+    repo.index.move([f'{tree}/exemplar', f'{tree}/{args.project_name}'])
+
+# Replace old README.md with new one.
+readme = f'''<!--
+SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+-->
+
+# beman.{args.project_name}: {args.desc}
+
+<img src="https://github.com/bemanproject/beman/blob/main/images/logos/beman_logo-beman_library_under_development.png" style="width:5%; height:auto;"> ![Continuous Integration Tests](https://github.com/{args.owner}/{args.project_name}/actions/workflows/ci_tests.yml/badge.svg) ![Code Format](https://github.com/{args.owner}/{args.project_name}/actions/workflows/pre-commit.yml/badge.svg)
+
+**Implements**: [`{args.project_name}` ({args.paper})](https://wg21.link/{args.paper})
+
+**Status**: [Under development and not yet ready for production use.](https://github.com/bemanproject/beman/blob/main/docs/BEMAN_LIBRARY_MATURITY_MODEL.md#under-development-and-not-yet-ready-for-production-use)
+
+TODO: Describe this project.
+
+### Usage
+
+```c++
+TODO: Copy/paste example code from examples/ here.
+```
+
+## Building beman.{args.project_name}
+
+### Dependencies
+This project has no C or C++ dependencies.
+
+Build-time dependencies:
+
+- `cmake`
+- `ninja`, `make`, or another CMake-supported build system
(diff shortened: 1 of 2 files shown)


steve-downey:
This is about a day and a half late for my latest free-time project. 
Thanks!

tzlaine:
> This is about a day and a half late for my latest free-time project. Thanks!

If I'm known for anything, it's my ti--

steve-downey (approved):
I  ran this, and it surprised me a bit modifying the repo in place and committing the changed version. That's a possible footgun? 
Also, possibly this should be in a scripts or bin directory so it's not an accident waiting to happen? 
Both of those are pretty minor, though.

new_project_from_exemplar.py
```diff
@@ -0,0 +1,277 @@
...
+<!-- TODO Darius: rewrite section!-->
+```shell
+# TODO
+```
+
+</details>
+
+<details>
+<summary>Dependencies install {args.project_name} on Windows $VERSION  </summary>
+<!-- TODO Darius: rewrite section!-->
```
neatudarius, line 90:
```suggestion
<!-- TODO: rewrite section!-->
```

new_project_from_exemplar.py
```diff
@@ -0,0 +1,277 @@
...
+  g++-14 gcc-14                           \
+  clang-18 clang++-18 clang-19 clang++-19
+```
+
+</details>
+
+<details>
+<summary>Dependencies install {args.project_name} on MAC OS $VERSION </summary>
+
+<!-- TODO Darius: rewrite section!-->
```
neatudarius, line 81:
```suggestion
<!-- TODO: rewrite section!-->
```

new_project_from_exemplar.py
```diff
@@ -0,0 +1,277 @@
...
+```cmake
+target_link_libraries(yourlib PUBLIC beman::{args.project_name})
+```
+
+</details>
+
+<details>
+<summary> Use beman.{args.project_name} from other build systems </summary>
+
+<!-- TODO Darius: rewrite section! Add examples. -->
```
neatudarius, line 204:
```suggestion
<!-- TODO: rewrite section! Add examples. -->
```

neatudarius (commented):
I think this PR is good. Note that is an overlap with https://github.com/bemanproject/infra/issues/1. TLDR: I think https://github.com/bemanproject/infra/issues/1 can superseed this PR.

tzlaine:
> I ran this, and it surprised me a bit modifying the repo in place and committing the changed version. That's a possible footgun? Also, possibly this should be in a scripts or bin directory so it's not an accident waiting to happen? Both of those are pretty minor, though.

That's the beauty of version control, no?  You run the thing once, it modifies stuff in place, and then removes itself.  But the state of the repo before that happened remains.  You can always go back.

new_project_from_exemplar.py
tzlaine: Lemme get this straight -- you don't want all future Beman projects to have TODOs for you to handle? :)

steve-downey:
We should probably just have the 'owner' in the CODEOWNERS file, too.

new_project_from_exemplar.py
tzlaine: Ok, this is now done (removal of "Darius" comments).

tzlaine:
> We should probably just have the 'owner' in the CODEOWNERS file, too.

Good call.  I didn't even know about that file.  Done.

wusatosi (commented):
Cool tool!

Though I think this is more appropriate for the infrastructure repo?

exemplar is mainly for creating a usable repo directly, I don't see why every repo created from exemplar should have a copy of this script.

tzlaine:
> Cool tool!
> 
> Though I think this is more appropriate for the infrastructure repo?
> 
> exemplar is mainly for creating a usable repo directly, I don't see why every repo created from exemplar should have a copy of this script.

I think it fits the intended use case quite well:

1) Click the 'Use this template' button.
2) Run the script; the script makes a commit including the changes, and then removes itself.
3) You proceed to write your code.
4) Details TBD ...
5) Profit!

If it were in a separate repo, this would not go as smoothly, IMO.  I have not yet investigated David Sankels' approach to automating this so that 1) and 2) above collapse into one step.  That seems good to.

steve-downey:
Add that to the README and delete it too?

tzlaine:
> Add that to the README and delete it too?

Ooh, that's a good idea!

wusatosi:
> > Cool tool!
> > Though I think this is more appropriate for the infrastructure repo?
> > exemplar is mainly for creating a usable repo directly, I don't see why every repo created from exemplar should have a copy of this script.
> 
> I think it fits the intended use case quite well:
> 
> 1. Click the 'Use this template' button.
> 2. Run the script; the script makes a commit including the changes, and then removes itself.
> 3. You proceed to write your code.
> 4. Details TBD ...
> 5. Profit!
> 
> If it were in a separate repo, this would not go as smoothly, IMO. I have not yet investigated David Sankels' approach to automating this so that 1) and 2) above collapse into one step. That seems good to.

Okay this makes sense for me, you're right. Can you add this instruction to the README so its more obvious?

tzlaine:
> Okay this makes sense for me, you're right. Can you add this instruction to the README so its more obvious?

Done.
 
> Also will we be able to test this? At least make sure this script will run without error?

Weeeeell, this one is a bit tricky.  I mean, the way I tested locally it could be automated, in that you could check everything out, then run the script, then build and run the tests.  Ok, that's not actually all that tricky, now that I write it out.

Hm, yeah.  I think this could be its own block of CI tests; that block could be marked, and removed from the CI YAML file by the script.

wusatosi:
> > Also will we be able to test this? At least make sure this script will run without error?
> 
> Weeeeell, this one is a bit tricky. I mean, the way I tested locally it could be automated, in that you could check everything out, then run the script, then build and run the tests. Ok, that's not actually all that tricky, now that I write it out.
> 
> Hm, yeah. I think this could be its own block of CI tests; that block could be marked, and removed from the CI YAML file by the script.

I can help out with the CI part. Can you provide a bash script that:
- Invokes the script with a preset combination of descriptions/ owners/ etc
- Do some kind of verification on "there's something that has been done", it does not need to be exhaustive and could be optional.
- return with 0 if the script ran without fault and the check passes, fails with non 0 exit code if the script failed or the check fails

Sorry this is tricky work, but I think we should be able to verify this script is somewhat valid, and the generate project is valid. Again I can take-care of the CI part.

steve-downey:
> Plus, if pre-commit is correctly installed, project author will not be able to commit unless the linting errors are addressed.

We should be very careful about anything that installs pre-commit hooks in the local copy. Committing partly linty code is an important workflow, especially if you want to separate meaningful human code from scripting fix-up. 

This is very much to the side of this work, though.

camio:
Any plans to continue moving this PR forward? If not, I'm leaning towards closing it since there hasn't been any progress in 4 months. We can always re-open it later.

neatudarius:
I think this tool should be moved inside beman.infra. @camio ?

wusatosi:
> I think this tool should be moved inside beman.infra.

I don't think this is being persuade, we should simply close this.

PR #94, opened 2025-01-03
cargo-generate project template POC

camio:
This is a proof of concept that turns exemplar into a template that
generates skeleton Beman projects using the [cargo-generate tool](https://github.com/cargo-generate/cargo-generate).

See https://github.com/camio/foobar.git for an example skeleton called
"foobar" generated from this template.

Use of this template requires a cargo-generate installation:

1. [Install Rust](https://www.rust-lang.org/tools/install) (a cargo-generate prerequisite)
2. `cargo install cargo-generate` (detailed instructions [here](https://cargo-generate.github.io/cargo-generate/installation.html))

Once `cargo-generate` is installed, run

```
cargo generate gh:camio/example --branch cargo-generate
```

and follow the prompts to generate a template.


wusatosi:
This seems really cool!

Though is there any similar tool that fall into the python eco-system? We already need python/ pip for pre-commit. It would be better if we not start pulling toolings from too many package managers.

neatudarius:
@wusatosi , do you plan to use this PoC for your template work?

PR #95, opened 2025-01-04
Bring into compliance with [CMAKE.SKIP_TESTS] section of Beman Standard

ednolan:
This commit replaces the use of BUILD_TESTING with a BEMAN_EXAMPLE_BUILD_TESTING option that is enabled by default iff the project is not being included by another project, as recommended by the Beman Standard.


ednolan:
Created this by accident due to forgetting to git pull my old checkout of the repository. This change has already been applied to exemplar on the latest version.

PR #101, opened 2025-01-16
Use parallel build in CI

wusatosi:
The GitHub Actions runner [has 4 cores](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for--private-repositories), parallel build should speed up the build process.

Unfortunately, there's not a portable way to add this in preset, see: [here](https://discourse.cmake.org/t/presets-use-number-of-processors-for-number-of-parallel-jobs/4989).

### DIFF
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -130,7 +130,7 @@ jobs:
       - name: Build Release
         run: |
           # Portable commands only
-          cmake --build build --config Release --verbose
+          cmake --build build --config Release --parallel --verbose
           cmake --build build --config Release --target all_verify_interface_header_sets
           cmake --install build --config Release --prefix /opt/beman.exemplar
           ls -R /opt/beman.exemplar
@@ -139,7 +139,7 @@ jobs:
       - name: Build Debug
         run: |
           # Portable commands only
-          cmake --build build --config Debug --verbose
+          cmake --build build --config Debug --parallel --verbose
           cmake --build build --config Debug --target all_verify_interface_header_sets
           cmake --install build --config Debug --prefix /opt/beman.exemplar
           ls -R /opt/beman.exemplar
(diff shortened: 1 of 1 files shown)


neatudarius:
What's the new duration? (old vs new)

dietmarkuehl:
Silly question on the CI: is there a way to run the CI (or parts thereof if things depened on the OS) locally?

.github/workflows/ci_tests.yml
```diff
@@ -117,7 +117,7 @@ jobs:
       - name: Build Release
```
neatudarius, line 117: Can we do parallel jobs ? e.g. `Build Release` || `Build Debug`

neatudarius (approved):

.github/workflows/ci_tests.yml
wusatosi: The only way to run these in parallel is to separate them into multiple jobs. This would bring speedup but it would make number of checks double.

wusatosi:
> Silly question on the CI: is there a way to run the CI (or parts thereof if things depened on the OS) locally?

Unfortunately GitHub Action is notoriously not testable, there's no officially supported execution engine, but [Ack](https://github.com/nektos/act) is a community maintained implementation that allows you to run the pipeline locally.

steve-downey:
> Silly question on the CI: is there a way to run the CI (or parts thereof if things depened on the OS) locally?

Not easily, unfortunately. I've had some success with https://github.com/nektos/act and when it works it works fine, but it's a bit tricky to get working.

wusatosi:
> Not easily, unfortunately. I've had some success with https://github.com/nektos/act and when it works it works fine, but it's a bit tricky to get working. That's the same as the project that @wusatosi mentioned.

Yeah I personally find creating a branch and putting up debugging commits to be, ironically, the easiest approach.

dietmarkuehl:
Thanks for the pointer to ACT: I'll have a look. My objevtive isn't to test the CI bit rather to test the code, assuming the CI works. I'll see how that goes.

PR #103, opened 2025-01-25
No fail fast on CI tests

wusatosi:
Right now, if one of the preset tests or compiler test reported error, the entire preset/ compile test pipeline will be cancelled.

See example for compiler here: https://github.com/bemanproject/lazy/actions/runs/12743334139/job/35513073020
For presets here: 
Presets here: https://github.com/bemanproject/inplace_vector/actions/runs/12960992490

Turning fail-fast off maybe more helpful, as it will make platform specific errors easier to identify.

Before: 
image

After:
image

This is already added to beman.lazy, see: https://github.com/bemanproject/lazy/commit/aba368a493945757f8a981304e0285fe3b69e5e1

### DIFF
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -143,6 +143,7 @@ jobs:
   configuration-test:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         args:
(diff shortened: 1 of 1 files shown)


bretbrownjr (approved):
Great idea!

PR #104, opened 2025-01-25
Support BUILD_SHARED_LIBS

bretbrownjr:
## Problem

Hardcoding STATIC is unneeded and breaks support for the standard CMake feature `BUILD_SHARED_LIBS`.

## Solution

Remove `STATIC` from the `add_library(exemplar)` statement and allow CMake defaults to provide a default static library build.

## Reference

For `add_library` supporting `BUILD_SHARED_LIBS`:
 https://cmake.org/cmake/help/latest/command/add_library.html

Docs for `BUILD_SHARED_LIBS`, including that `STATIC` is the default for CMake built libraries.
 https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html#variable:BUILD_SHARED_LIBS

### DIFF
--- a/src/beman/exemplar/CMakeLists.txt
+++ b/src/beman/exemplar/CMakeLists.txt
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-add_library(beman.exemplar STATIC)
+add_library(beman.exemplar)
 add_library(beman::exemplar ALIAS beman.exemplar)
 
 target_sources(beman.exemplar PRIVATE identity.cpp)


bretbrownjr:
Hardcoding `STATIC` in `add_library` commands is a way to explicitly remove support for shared library builds, though I don't believe we forbid or even discourage shared library linking in the Beman Standard at this point. Presumably some libraries would like to support this feature, especially because standard C++ library implementations do release as shared libraries.

wusatosi (approved):
Seems like a very reasonable update for me.

neatudarius (approved):
LGTM 👍 

Please open an issue here https://github.com/bemanproject/beman/issues. We need to update the Beman Standard and all repos.

PR #105, opened 2025-01-26
Inject FetchContent logic

bretbrownjr:
Problem
-------

**Given**:

A sandbox environment is an environment where dependencies are all provided but access to the internet is not. Also take as a given that the build for this project happens in an environment that meets that description.

**When**:

Configuring this project with any of the provided approaches. For instance:

```
cmake -B build -S .
```

**Then**:

The configuration step hangs or fails, depending on the nature of the sandboxed environment.

Solution
--------

Move to a `find_package` first implementation for the `CMakeLists.txt` of this project.

Using `FetchContent` is still trivial, using the new functionality in `cmake/use-fetch-content.cmake` as documented in the updates to `README.md` included in this commit.


bretbrownjr:
This is an alternative and mutually exclusive approach to #92.

Note that it uses yet another approach to letting the same repo use either `FetchContent` or `find_package` without hardcoding anything special in `CMakeLists.txt`. See https://cmake.org/cmake/help/latest/guide/using-dependencies/index.html#id11 and the documents it links for more on how "Dependency Providers" work.

This approach is more code compared to #92, but I would expect if we liked this approach, the entirety of the `use-fetch-content.cmake` file could spun off to a standalong utility project, perhaps inside https://github.com/bemanproject/infra. This seems to be an ergonomic gap in CMake, so analogous features could be proposed upstream to CMake itself assuming we get a usage experience we're happy with.

Another merit of this approach compared to #92 is that pinned dependency information is in legible JSON format, so automation (like a `make update` target) to update the pinned version of GoogleTest would be very possible compared to hardcoding that information in a `CMakeLists.txt`. Similarly, documentation would be relatively simpler and require no special CMake knowledge.

bretbrownjr:
On the failing CI checks, they are due to a missing googletest dependency, which isn't a large concern. It's not too difficult to have presets that prefer `FetchContent` on MacOS and Windows, pending selection of package management solutions on those architectures. Same for installing `libgtest-dev` on Ubuntu.

First, I'd like to hear whether this is a design we would like to pursue before investing time in troubleshooting CI failures.

wusatosi (commented):
Is this too much complexity for supporting using local GTest package? Looks like you are trying to implement a package manager.

bretbrownjr:
@wusatosi Using `FetchContent` is implementing dependency management in-repo, yes.

Using it explicitly is an incomplete approach to dependency management since it doesn't have the features (sandbox-friendliness, portability to existing packaging systems) that I'm looking for with this PR.

This PR demonstrates how it might be injected. I don't expect this is a good long-term approach. As [I state in the Discourse on this change](
https://discourse.bemanproject.org/t/make-beman-packageable/331), if we take this direction as a medium-term strategy, I expect the interesting CMake logic would be pulled out into another repo like https://github.com/bemanproject/infra.

Another approach to consider would be to [support Conan with a `conanfile.txt`](https://docs.conan.io/2/tutorial/consuming_packages/build_simple_cmake_project.html) and/or [support vcpkg with a `vcpkg.json`](https://learn.microsoft.com/en-us/vcpkg/consume/manifest-mode?tabs=cmake%2Cbuild-cmake) directly in this repo. One of both of those approaches could be used in CI, assumed in presets, etc.

`FetchContent` would still be possible, but it would be an exercise for consumers of this project to wire that up, perhaps with the same `CMAKE_PROJECT_TOP_LEVEL_INCLUDES` and dependency provider approach if they desired.

bretbrownjr:
This latest commit fixes the linting checks, especially CMake formatting issues.

The remaining CI failures revolve around actually installing `GTest`. Those are going to require a consensus decision on how to resolve. Look out for another draft PR by me to demonstrate how to resolve that issue with a `conanfile.txt`.

cmake/use-fetch-content.cmake
```diff
@@ -0,0 +1,175 @@
...
+    BASE_DIRECTORY "${BemanExemplar_projectDir}"
+    EXPAND_TILDE
+)
+message(DEBUG "Using lockfile: \"${BemanExemplar_lockfile}\"")
+
+# Force CMake to reconfigure the project if the lockfile changes
+set_property(
+    DIRECTORY "${BemanExemplar_projectDir}"
+    APPEND
+    PROPERTY CMAKE_CONFIGURE_DEPENDS "${BemanExemplar_lockfile}"
```
chayden83, line 28: I didn't know about `CMAKE_CONFIGURE_DEPENDS`. Interesting!

bretbrownjr:
I made a PR against the Beman Standard to land alongside this change that recommends `find_package` instead of `FetchContent`:
https://github.com/bemanproject/beman/pull/98

bretbrownjr:
Forgive the CI-driven-development, but the HEAD commit on this PR now works.

This is ready for review, and if acceptable, merging.

.github/workflows/ci_tests.yml
```diff
@@ -47,7 +47,11 @@ jobs:
         with:
           arch: x64
       - name: Run preset
-        run: cmake --workflow --preset ${{ matrix.presets.preset }}
+        run: |
+          cmake -DCMAKE_PROJECT_TOP_LEVEL_INCLUDES="./cmake/use-fetch-content.cmake" --preset ${{ matrix.presets.preset }}
+          cmake --build --preset ${{ matrix.presets.preset }} --parallel --verbose
+          cmake --build --preset ${{ matrix.presets.preset }} --parallel --verbose --target all_verify_interface_header_sets
+          ctest --preset ${{ matrix.presets.preset }}
```
bretbrownjr, line 54: Allow me to explain this change as it is required given the current design.

`cmake --workflow --preset ....` no longer supports FetchContent.
This is because `--workflow` does not allow us to provide `-D` flags to tweak the workflow.
The `-D` flag is required to allow us to select between `FetchContent` or
`find_package` logic at CMake configure time. Given there is not option to make
that choice, the `find_package` logic is the only approach supported by `--workflow`
right now.

That being said, I could probably come up with a design involving injecting
that logic in a toolchain file and using environment variables to toggle the
choice, but I'd rather not implement something like that without feedback from
the community.

At any rate, note that the CI that used to use `--workflow` instead uses the
multi-command workflow with the preset configuration selected by the CI
matrix.

.github/workflows/ci_tests.yml
```diff
@@ -97,7 +101,10 @@ jobs:
             cmake_args:
               description: "TSan"
 
-    name: "Unit: ${{ matrix.platform.description }} ${{ matrix.cpp_version }} ${{ matrix.cmake_args.description }}"
+    name: "Unit:
+              ${{ matrix.platform.description }}
+              ${{ matrix.cpp_version }}
+              ${{ matrix.cmake_args.description }}"
```
bretbrownjr, line 107: I edited this line and a few others like it to make the file easier to read, diff, and review. Hopefully everyone's OK with doing that as part of this PR.

.github/workflows/ci_tests.yml
```diff
@@ -114,15 +121,20 @@ jobs:
...
           ninja --version
       - name: Configure CMake
-        run: |
-          cmake -B build -S . -DCMAKE_CXX_STANDARD=${{ matrix.cpp_version }} -DCMAKE_TOOLCHAIN_FILE="${{ matrix.platform.toolchain }}" ${{ matrix.cmake_args.args }}
+        run: 'cmake
+                 -B build
+                 -S .
+                 -DCMAKE_CXX_STANDARD=${{ matrix.cpp_version }}
+                 -DCMAKE_TOOLCHAIN_FILE="${{ matrix.platform.toolchain }}"
+                 -DCMAKE_PROJECT_TOP_LEVEL_INCLUDES="./cmake/use-fetch-content.cmake"
```
bretbrownjr, line 136: Let me note since the formatting changes obscure the change a little. This argument referencing `use-fetch-content.cmake` is added. It makes this CI workflow use `FetchContent`.

Pay attention to the other `cmake` commands in this file as well. If you do not see a mention of `use-fetch-content.cmake`, the workflow uses `find_package` and system installed dependencies.

wusatosi (changes requested):
## Config priority

Do we really want searching for local `libgtest` to be the default behavior?

For people hacking around beman or downstream paper/implementation authors, I don't think they will be more likely to care about offline run with `libgtest` installed than being able configure the project with one click.

I think the priority is off, it should be:
First: People with generic environment (cmake, gcc, a network connection and no libgtest) should have the ability to configure the project with one click.
Second: People that want to configure the project for offline configuration (e.g. they are so interested in the project that they bring the project on a plane or into their company) should have the option via an explicit parameter to do an offline build.

## Need docs for downstream library authors

Missing docs for downstream library author that want to add a beman project (e.g. iterator-interface or execution) as sub-dependency. Do they have to do `-DCMAKE_PROJECT_TOP_LEVEL_INCLUDES="./cmake/use-fetch-content.cmake"` all the time? 

The internals of `lockfile.json` is not documented (why not `dependencies.json` btw), and it's not clear how offline functionality will interact with them. I think this is significant enough of a change to require documentation before merging.

## Including fetch content shouldn't look like magic

I would prefer if we can switch this behavior with something like: `BEMAN_DEPENDENCY_FETCH=LOCAL` (as enum option with `local`, `fetch`. so we can add maybe `vcpkg` in the future).

This would be a lot more declarative and obvious to change for downstream than adding: `CMAKE_PROJECT_TOP_LEVEL_INCLUDES="./cmake/use-fetch-content.cmake`.

.github/workflows/ci_tests.yml
wusatosi: I feel like the fact this much change is needed for testing preset is confirming to my "inversed priority" critic. The default behavior should be `fetchcontent` but not `apt install libgtest-dev`. If we cannot configure the project using a preset in one of the most developer tool bloated environments that is the default GitHub actions image, we are not easy to setup.

.github/workflows/ci_tests.yml
```diff
@@ -162,13 +174,18 @@ jobs:
...
           cmakeVersion: "~3.25.0"
           ninjaVersion: "^1.11.1"
-      - name: Print installed softwares
+      - name: Print installed software
         run: |
           cmake --version
           ninja --version
       - name: Configure CMake
         run: |
-          cmake -B build -S . -DCMAKE_CXX_STANDARD=17 ${{ matrix.args.arg }}
```
wusatosi, line 171: Thanks : D ! , I don't know why I haven't done this.

cmake/use-fetch-content.cmake
```diff
@@ -0,0 +1,175 @@
+cmake_minimum_required(VERSION 3.24)
+
+if(NOT BEMAN_EXEMPLAR_LOCKFILE)
+    set(BEMAN_EXEMPLAR_LOCKFILE
```
wusatosi, line 4: Why not use options here? Is it because CACHE FILE PATH?

bretbrownjr:
> Do we really want searching for local libgtest to be the default behavior?

We want `find_package(GTest)` to be the way we declare an external dependency, yes. This is the recommended and portable practice as discussed. `FetchContent` does not provide the portability or ergonomics we'd want to replicate across the other Beman libraries and the C++ ecosystem more broadly.

> I would prefer if we can switch this behavior with something like: BEMAN_DEPENDENCY_FETCH=LOCAL (as enum option with local, fetch. so we can add maybe vcpkg in the future).

I'm opposed to implementing bits and pieces of a dependency management system in every Beman library. It makes more sense to move the whole community to supported and widely adopted technologies like Conan, vcpkg, and/or a number of other systems that already support fetching dependencies from the internet (or not) in all sorts of configurations.

I'll be clear about this. Downloading dependencies from github.com is supported by *both* Conan and vcpkg (and all the other options too!). It's really easy and covered by the introductory tutorials of both options. They both support building against a local cloned and modified of the `googletest` repo as well. We wouldn't have to develop and maintain tools and paragraphs of docs, just point our users at the most relevant sections, if that's required.

> I would prefer if we can switch this behavior with something like: BEMAN_DEPENDENCY_FETCH=LOCAL (as enum option with local, fetch. so we can add maybe vcpkg in the future).

A `BEMAN_DEPENDENCY_FETCH=conan` or `vcpkg` wouldn't really work. Conan is already running a command like `conan install` before *it* invokes `cmake`. vpkg is incorporated via a CMake toolchain file named `vcpkg.cmake`.

One of the biggest problems with `FetchContent` is that it encourages conflation of build systems and dependency management systems. This wouldn't be too bad except for the fact that the dependency management features it realistically provides does not stack up well at all compared to alternatives (see the discussion below about lockfiles).

I also don't want users to type `use-fetch-content.cmake` long term. This is an intermediate technology until we come up with something better. The main goal of adding it in this PR is to provide some sort of stopgap and to demonstrate that we can make `find_package(GTest)` resolve to... well, whatever we want.

> The internals of lockfile.json is not documented...

That's fair, and I can update the PR quickly to add some docs, though they are more developer documentation for maintainers of Exemplar than something I expect users will need to interact with. Meaning I'm not sure if I'll put them in a user-facing `README.md` or not.

I'm fairly flexible on the name, for what it's worth. If we can get consensus but only if the file is named `examplar.lock.json` or `fetchcontent.dependencies.json` or something, I can live with that.

To explain my reasoning, I chose the name "lockfile" because that's one term of art for files like these. If you ask your search engine or your favorite programmer-oriented LLM tool will turn up full explanations, I expect, but here's [the Conan Lockfile documentation](https://docs.conan.io/2/tutorial/versioning/lockfiles.html) for a quick and relevant example 

PS. By linking the Conan docs on the subject of dependency pinning, I'm throwing shade at `FetchContent` because Conan supports (and documents!) all sorts of dependency manipulation workflows and features that `FetchContent` really doesn't. And Conan is actually not exceptional in supporting this. On the contrary, `FetchContent` is notable for its limited built-in feature set.

wusatosi:
Cool! I don't and can't claim to know enough about CMake or package managers to provide more judgement, but as long as my concerns are seen I am happy to sit aside and learn by observation.

My only concern would be loss of one-click setup, I think a cheaty way would be to do this would be adding `CMAKE_PROJECT_TOP_LEVEL_INCLUDES=./cmake/use-fetch-content.cmake` to current CMake presets, and having presets that deal specificity with package management. Agh, only if you can pass arguments when using preset, ~~how much do we need to pay kitware to have this implemented?~~

cmake/use-fetch-content.cmake
mguid65: option is only intended for boolean configs

wusatosi: thanks for the context!

bretbrownjr:
> My only concern would be loss of one-click setup, I think a cheaty way would be to do this would be adding CMAKE_PROJECT_TOP_LEVEL_INCLUDES=./cmake/use-fetch-content.cmake to current CMake presets, and having presets that deal specificity with package management.

I think ergonomics are always a valid concern. Can you add an issue to track your concerns and we can address in future PRs? I would like to work on other features in parallel -- like Conan and vcpkg support.

On the specific idea, it's possible to add `CMAKE_PROJECT_TOP_LEVEL_INCLUDES`, yes. Though I do want to keep an eye on how many presets we support. Each adds more complexity to each Beman library, so there's a `O(n)` cost in terms of config file complexity. Each also requires more CI coverage, which is arguably `O(m*n)` or worse, considering we have other options: package managers, sanitizers, compilers, etc.,

bretbrownjr:
This PR is updated and ready for final review. If it's basically OK, I would like to merge so I can push up pull requests to initially support Conan and vcpkg. I'm happy to collaborate on followon PRs to provide more polish, iterate on how we inject FetchContent, etc.

@wusatosi I added some documentation for `lockfile.json` as requested.

README.md
```diff
@@ -175,6 +172,18 @@ as per [the official LLVM install guide](https://apt.llvm.org/).
...
+required. Here is an example of installing GoogleTest on a Debian-based Linux
+environment:
+
+```bash
+apt install libgtest-dev
+```
+
+The precise command and package name will vary depending on the Linux OS you are
+using. Be sure to consult documentation and the package repository for the system
+you are using.
```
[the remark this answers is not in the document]
wusatosi: Seems like we need to be based on Google Test `v1.16.0`, see #128 . `libgtest-dev` for latest LTS is [`1.14.0`](https://packages.ubuntu.com/noble/libgtest-dev), we need to document this version requirement and advise install gtest from source.

wusatosi (commented):
CI and docs looks good to me with minor suggestions.

This is a CMake heavy PR that I am not qualified to approve. I think @camio should take a crack at this.

I am still bit salty about the loss of ergonomics here, but I guess we can bring it back later.

I encourage reviewer to read through my comments and @bretbrownjr 's reply, there are significant decisions made in this PR.

README.md
```diff
@@ -88,16 +88,13 @@ This project requires at least the following to build:
...
-This project pulls [Google Test](https://github.com/google/googletest)
-from GitHub as a development dependency for its testing framework,
-thus requiring an active internet connection to configure.
-You can disable this behavior by setting cmake option
+You can disable building tests by setting cmake option
 [`BEMAN_EXEMPLAR_BUILD_TESTS`](#beman_exemplar_build_tests) to `OFF`
 when configuring the project.
 
-However,
-some examples and tests will not be compiled
```
wusatosi, line 100: Screenshot from 2025-03-02 02-22-53

Here's the overdue reminder... I think you've done what you need to be reminded about.

wusatosi:
Can you add this patch so the devcontainer still work?

```patch
diff --git a/.devcontainer/postcreate.sh b/.devcontainer/postcreate.sh
index bddba03..81d2420 100644
--- a/.devcontainer/postcreate.sh
+++ b/.devcontainer/postcreate.sh
@@ -1,3 +1,6 @@
 # Setup pre-commit
 pre-commit
 pre-commit install
+
+# Install GoogleTest
+sudo apt-get -y install libgtest-dev
```

docs/README.md
```diff
@@ -4,4 +4,80 @@ TODO: Currently only a skeleton.
...
+
+* `package_name` *must* match the upstream-documented "package name" that would
+  be provided to a `find_package` call. For GoogleTest, this is `GTest`, for instance.
+
+* `git_repository` is a full https URL for the repository to clone.
+
+* `git_tag` must be a valid git ref in that repository. This identifies precisely which
+  version of the dependency to build. While branch and tag names will work for this value,
+  the Beman Project prefers the stability provided by a full-length git commit ID, so
+  please use one of those in any changes submitted to `lockfile.json`.
```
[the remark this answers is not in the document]
bretbrownjr: It's a little hypothetical in this case, but if that helps, it's not hard to write that up.

README.md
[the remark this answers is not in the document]
bretbrownjr: I don't have a good development environment for MacOS at the moment. I would like to see a PR following this one to add support for that, both to docs and to the CI. At the moment, CI for MacOS uses FetchContent.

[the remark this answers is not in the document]
bretbrownjr: We only need GoogleTest `v1.16.0` to support `FetchContent` workflows. Given the simplicity of the tests as written, we can probably support GoogleTest going back a decade or more for packaging workflows.

README.md
bretbrownjr: Thanks for making sure we don't forget this detail!

bretbrownjr:
@wusatosi I added more docs walking through how to add a Catch2 dependency. Does that help?

I also added `libgtest-dev` to the devcontainer JSON. I did a little tidying up in there as well. Hopefully that's OK.

wusatosi (commented):
👌 no more complaints from me

README.md
```diff
@@ -175,6 +172,18 @@ as per [the official LLVM install guide](https://apt.llvm.org/).
 bash -c "$(wget -O - https://apt.llvm.org/llvm.sh)"
 ```
 
+If the included test suite is being built and run, a GoogleTest library will be
+required. Here is an example of installing GoogleTest on a Debian-based Linux
+environment:
+
+```bash
+apt install libgtest-dev
```
JeffGarland, line 180: I guess google can't follow established practice?  A -dev install is normally when you want to modify the library, not just use it.  But apparently this is the correct way to install gtest, sadly.

docs/README.md
```diff
@@ -4,4 +4,111 @@ TODO: Currently only a skeleton.
...
+
+#### Design
+
+This is a design for defining dependency providers
+[discussed in CMake upstream documentation][dependency-providers]. The
+`use-fetch-content.cmake` file *also* leverages CMake support for parsing
+JSON to get the details of projects to provide from `lockfile.json`. This:
+
+* Ensures that calls to FetchContent APIs within this project are consistent
+  and meet Beman Standards
```
JeffGarland, line 74:
```suggestion
  and meet Beman Standards.
```

docs/README.md
```diff
@@ -4,4 +4,111 @@ TODO: Currently only a skeleton.
...
+#### JSON Structure
+
+The `lockfile.json` file contains an object with one field named `dependencies`.
+
+`dependencies` should have a value that is an array of objects.
+
+Each dependency object should contain exactly four fields with string values:
+
+* `name` is used as the FetchContent name for the project. See
+  [the API docs for FetchContent][fetch-content] for more on what a "FetchContent
```
JeffGarland, line 95: is this supposed to be a link?

README.md
```diff
@@ -88,16 +88,13 @@ This project requires at least the following to build:
 
 * C++17
```
JeffGarland, line 89:
```suggestion
* C++20
```

Below we're now defaulting to 20, which seems to be the right default - suggest we update here too.

JeffGarland (approved):
Overall this looks like a good step forward -- I've made a couple small suggestions, but overall I think we should merge and improve from here. Approving.

README.md
bretbrownjr: Don't worry. This is the normal Debian/Ubuntu/etc. naming convention. `libfoo-dev` contains headers, static libraries, and anything else needed in a build environment to build against the `foo` library. `libfoo` would contain runtime dependencies for the `foo` library, but when statically linking, you generally don't have any of those.

See:
https://packages.ubuntu.com/oracular/libgtest-dev

Also, like most Google-originating projects, the packaging is maintained separately from the original library. You can see the maintainers listed in the above link. They're a combination of Debian and Ubuntu contributors. They get all the credit and blame for the Debian packaging details, I suppose.

README.md
bretbrownjr: I didn't change this detail. for what it's worth.

I believe the library supports C++17 entirely, but certain test cases require C++20 to run.

I'm pretty sure we could backport the library to C++03, even. I think all of the build requirements are related to the unit tests, but I've been focusing mostly on tooling details, so I haven't followed up on this detail.

Given you've approved the PR, I'll assume that means we're OK addressing the C++ standard compatibility in a separate PR or issue.

wusatosi: Bret is right here.

mguid65: This shouldn't affect this PR, but I was writing a convenience utility for processing json in cmake in a more declarative way instead of the convoluted string function that currently exists. Maybe unnecessary if we don't expect to write too much of this cmake code like this

neatudarius (approved):

bretbrownjr:
I'm merging this change given the approving reviews. @wusatosi, I'm happy to team up on follow up on issues and PRs addressing ergonomic concerns.

PR #106, opened 2025-01-30
Fix codespace ASAN stall

### LINKED ISSUE #102: ASAN is flaky on Docker, and our fix is a no-op.
To reproduce: In a Devcontainer environment (e.g. GitHub Codespaces), run

```bash
cmake --workflow --preset gcc-debug
```

Notice that the test suite is flaky - it will hang some of the time.

[Our current solution](https://github.com/bemanproject/exemplar/blob/main/.devcontainer/Dockerfile#L13-L17) is moving in the right direction, but sadly Docker containers silently fail to update this setting (for security reasons). Some `sysctl` options are available, but the one we need (`vm.mmap_rnd_bits`) is not. Note that moving the command to `postcreate.sh` also fails silently.

The correct solution either requires:
- Docker support for `docker run --sysctl vm.mmap_rnd_bits` (unlikely to arrive soon)
- Updating our LLVM version to one which has [the fix](https://github.com/llvm/llvm-project/commit/0784b1eefa36d4acbb0dacd2d18796e26313b6c5).

It is unclear if there are ASAN options we can pass to mitigate this issue in the interim.

wusatosi:
This fix #102 . Go read #102 for more of a write-up (Thanks nick!).

This PR updates codespace to use gcc-14.

Please test this PR by going to codespace and type `cmake --workflow --preset gcc-debug`, codespace should run the preset without fail/ hang.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,10 +1,10 @@
-FROM mcr.microsoft.com/devcontainers/cpp:1-ubuntu-22.04
+FROM mcr.microsoft.com/devcontainers/cpp:1-ubuntu-24.04
 
 USER vscode
 
 # Install latest cmake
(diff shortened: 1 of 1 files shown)


steve-downey (approved):
Looks like what I would expect for 24.04

PR #108, opened 2025-01-30
Error message when provided environment is below C++17

### LINKED ISSUE #40: Give user a better error message when the selected compiler is missing required functionality
When the user selects a compiler+flags combination that is insufficient to build the project, they're left with a build error  that provides no hint that the compiler+flags combination is the problem. This frequently comes up when a user doesn't use `-DCMAKE_CXX_STANDARD` when invoking CMake. See https://github.com/beman-project/exemplar/pull/38 and https://github.com/beman-project/inplace_vector/pull/6 for examples of this.

@ComicSansMS and @DeveloperPaul123 suggested using [cmake-compile-features](https://cmake.org/cmake/help/latest/manual/cmake-compile-features.7.html) with language standard arguments, but this has the side-effect of modifying the compiler flags selected by the user running CMake which can create incompatibilities.

This task is to incorporate a `try_compile` in CMakeLists.txt that attempts to compile a C++ file with the required language features. If it fails, a CMake-time error can be emitted that suggests the user use `-DCMAKE_CXX_STANDARD`. For beman.example, try_compile will verify `namespace foo::bar` compatibility.

wusatosi:
This closes #40 .

The update at `examples` directory may not be needed.

Output when C++11 is configured:
```
⋊> ~/b/exemplar on error_message ◦ cmake -S . -B build -DCMAKE_CXX_STANDARD=11
-- The CXX compiler identification is Clang 18.1.8
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Performing Test COMPILER_SUPPORTS_CPP_17
-- Performing Test COMPILER_SUPPORTS_CPP_17 - Failed
CMake Error at CMakeLists.txt:46 (message):
  Your compiler isn't advertising support for C++17 with the current
  configuration.Try adding -DCMAKE_CXX_STANDARD=17 to your CMake
  invocation.To override this check, set
  BEMAN_EXEMPLAR_OVERRIDE_CPP_17_REQUIREMENT to ON.


-- Configuring incomplete, errors occurred!
```

Specifically requesting review from @camio and @bretbrownjr .


wusatosi:
Agh! MSVC!!!!

neatudarius (approved):

CMakeLists.txt
```diff
@@ -11,6 +11,9 @@ project(
 
 enable_testing()
 
+include(cmake/feature_test.cmake)
```
bretbrownjr, line 14: @neatudarius Do you expect we'll be moving utilities like this into https://github.com/bemanproject/infra at some point? Or should we spin out another project for reusable CMake bits?

cmake/feature_test.cmake
```diff
@@ -0,0 +1,45 @@
...
+    int main() {}
+    #else
+    #error
+    #endif
+    "
+        ${RESULT_VAR}
+    )
+endfunction()
+
+function(BEMAN_MIN_CXX_VERSION CXX_VERSION)
```
bretbrownjr, line 16: I'm ok merging names like this, but keep for future reference, it's good to pick a "module name" and be consistent about prefixing any function names, variables, or anything else that might collide with other CMake entities anywhere in the CMake ecosystem. See here:
https://gitlab.kitware.com/cmake/cmake/-/blob/master/Help/dev/documentation.rst#module-functions-and-macros

No changes necessary right now because I'm pretty sure we'll end up turning this into a packaged module at some point anyway. If it's named `Beman`, then no name changes are needed.

cmake/feature_test.cmake
```diff
@@ -0,0 +1,45 @@
...
+        message(FATAL_ERROR "Unkown required C++ version: ${CXX_VERSION}")
+    else()
+        math(EXPR INDEX "${INDEX} + 1")
+        list(GET VERSION_MAP ${INDEX} VERSION_DATE)
+    endif()
+
+    beman_feature_test("__cplusplus >= ${VERSION_DATE}" CXX_VERSION_MET)
+    if(NOT CXX_VERSION_MET)
+        message(
+            FATAL_ERROR
```
bretbrownjr, line 41: We can merge this as is and see if anyone complains... but I could see someone wanting to have an option to override our CMake helper and try building anyway. We could provide an CMake variable of some sort that would turn this message into a less severe category like `WARNING` or even `NOTICE`. See all the options in the docs for the CMake message command:

https://cmake.org/cmake/help/latest/command/message.html

bretbrownjr (changes requested):
Excellent PR!

I'm happy to see this merged after fixing the typo I found.

CMakeLists.txt
wusatosi: I am designing this so it would be easily packagable into a CMake module. I would prefer putting this into infra for now.

cmake/feature_test.cmake
```diff
@@ -0,0 +1,45 @@
...
+        201703L
+        20
+        202002L
+        23
+        202302L
+    )
+
+    list(FIND VERSION_MAP "${CXX_VERSION}" INDEX)
+    if(INDEX EQUAL -1)
+        message(FATAL_ERROR "Unkown required C++ version: ${CXX_VERSION}")
```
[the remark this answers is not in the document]
wusatosi: Agh thanks!!

cmake/feature_test.cmake
wusatosi: This is exemplar, so we shouldn't rush anything. I would rather we get this right than we rush this in.

Right now, we have an internal arg used explicitly by CI and presets named `BEMAN_BUILDSYS_SANITIZER`.

Would it be a good idea to fold everything into this `BEMAN_BUILDSYS` name space?

Or would it be better called something like `BEMAN_FEATURE_TEST_ENFORCE_MIN_CXX`?

cmake/feature_test.cmake
wusatosi: Egh, that's probably a good idea especially considering detecting this on MSVC is problematic.

cmake/msvc-toolchain.cmake
```diff
@@ -26,10 +26,14 @@ if(BEMAN_BUILDSYS_SANITIZER STREQUAL "MaxSan")
...
-set(CMAKE_C_FLAGS_DEBUG_INIT "/EHsc /permissive- ${SANITIZER_FLAGS}")
+# /Zc:__cplusplus is needed for us to detect C++ version support portably
+# See: https://stackoverflow.com/questions/60052108/how-to-check-c-version-in-microsoft-visual-studio-2017
+set(BASE_FLAGS "/Zc:__cplusplus /EHsc /permissive-")
 
-set(RELEASE_FLAGS "/EHsc /permissive- /O2 ${SANITIZER_FLAGS}")
+set(CMAKE_CXX_FLAGS_DEBUG_INIT "${BASE_FLAGS} ${SANITIZER_FLAGS}")
+set(CMAKE_C_FLAGS_DEBUG_INIT "${BASE_FLAGS} ${SANITIZER_FLAGS}")
+
+set(RELEASE_FLAGS "/O2 ${BASE_FLAGS} ${SANITIZER_FLAGS}")
```
wusatosi, line 37: Hey @bretbrownjr I want to get your $0.02 on this.

Basically the `__cplusplus` feature flag detection is not universal on MSVC (agh! its always MSVC), you have to set this special flag `/Zc:__cplusplus` for MSVC to run correctly.

You technically could do a lot more complicated feature flag testing with if this is MSVC and if it is we test for `_MSVC_LANG` instead of `__cplusplus` etc... etc... so the standard library test would be portable, but it would be a pain.

So right now the approach is to have this flag set in our toolchain file (so it gets set correctly when using preset), but if someone configure this project without using preset, our detection could fail. This could potentially mean someone on MSVC couldn't configure the project at all if they don't put the flag.

Is this okay?

cmake/feature_test.cmake
```diff
@@ -0,0 +1,54 @@
...
+    beman_feature_test("__cplusplus >= ${VERSION_DATE}" CXX_VERSION_MET)
+    if(NOT CXX_VERSION_MET)
+        if(MSVC)
+            message(
+                WARNING
+                "Note: C++ version support detection requires a well-formed __cplusplus feature flag, \
+            MSVC does not set this flag correctly unless /Zc:__cplusplus flag is passed through.\
+            If you believe minimum C++ version detection is function incorrectly,\
+            please double check your compiler flags or use the provided presets if available."
+            )
```
wusatosi, line 47: @bretbrownjr is this a good enough hint message?, see the annotation on the toolchain file regarding MSVC's problem with `__cplusplus`.

wusatosi:
```suggestion
            message(
                WARNING
                "Note: C++ version support detection requires a well-formed __cplusplus feature flag, \
            MSVC does not set this flag correctly unless /Zc:__cplusplus flag is passed through.\
            If you believe minimum C++ version detection is functioning incorrectly,\
            please double check your compiler flags or use the provided presets if available."
            )
```

cmake/feature_test.cmake
```diff
@@ -0,0 +1,54 @@
+include(CheckCXXSourceCompiles)
+
```
camio, line 2: There isn't a good portable way to determine whether a compiler supports a given C++ revision's features. The generalization here isn't very useful.

I think this file should have a single function with this signature:

```cmake
# Determines if the selected C++ compiler has C++ 17 support. Sets
# 'result_var' to whether support is detected.
function(beman_exemplar_check_cpp_17_support result_var)
```

This can be implemented with a single `check_cxx_source_compiles` that first checks `__cplusplus > 201703L`. If that fails, see if `_MSC_VER` is defined. If it is, you can check for a couple C++17 feature flags.

Note that I prefixed the function with `beman_exemplar` so it won't conflict with other functions in a larger project.

There should also be a way to override the failure at CMake time in case `beman_exemplar_check_cpp_17` results in a false negative. In the top-level `CMakeLists.txt` file, there should be something like this:

```cmake
beman_exemplar_check_cpp_17_support(COMPILER_SUPPORTS_CPP_17)

option(
  BEMAN_EXEMPLAR_OVERRIDE_CPP_17_REQUIREMENT
  "Attempt building even though the compiler doesn't advertise the requisite C++17 support"
  OFF
)

if(NOT BEMAN_EXEMPLAR_OVERRIDE_CPP_17_REQUIREMENT AND NOT COMPILER_SUPPORTS_CPP_17)
  message(
    FATAL_ERROR
      "Your compiler isn't advertising support for C++17 with the current configuration. Try adding -DCMAKE_CXX_STANDARD=17 to your CMake invocation. To override this check, set BEMAN_EXEMPLAR_OVERRIDE_CPP_17_REQUIREMENT to ON."
  )
endif()
```

wusatosi: I am not a fan of this, I think we would eventually move this version detection functionality out of exemplar to the infra repo, and handle detection over there.

There is already significant friction on using exemplar due to the amount of `exemplar` one needs to replace across CMake files, I don't really want to add an extra reference to check other than
1. CMakeLists.txt
2. src/beman/.../CMakeLists.txt
3. tests/beman/.../CMakeLists.txt
4. examples/CMakeLists.txt

Would adding:
```CMake
BEMAN_FEATURE_TEST("__cplusplus > 201703L" COMPILER_SUPPORTS_CPP_17)
if (NOT COMPILER_SUPPORTS_CPP_17)
    message("BOOM")
endif()
```
To the main `CMakeLists.txt` and enforce `/Zc:__cplusplus` on MSVC family of compiler be a good enough part-in-the-middle?

CMakeLists.txt
bretbrownjr: Sounds good. I agree.

cmake/feature_test.cmake
bretbrownjr: I think a `BEMAN_` prefix is enough for now. I was just noting the best practice. If we make a function, variable, option, or any other CMake "name" specific to a project, the "name" must also be somehow specific to the project. Everyone that's not me help keep an eye on that because it's a *real* pain to clean up later when someone finally notices that `FetchContent` doesn't work because everyone's using the same `BUILD_INTEGRATION_TESTS` variable or something.

cmake/feature_test.cmake
bretbrownjr: Should we just admit it's time to make a https://github.com/bemanproject/cmake-beman that supports a `find_package(Beman REQUIRED)` feature to define these sorts of utilities?

If we're OK with some jank (git submodules or FetchContent or something), it's not too hard for me to crank out a basic repo like that in an afternoon.

wusatosi: Git submodules are discouraged in the beman standard.

Again I don't know how I feel about CMake configuration itself having a dependency, but I think this will definitely make beman rules like passive projects better more easily approachable as we provide beman variance.

chayden83: I think at some point, a Beman project devoted to shared CMake configuration is inevitable. If we agree on that, then I think we should make a project for it, as @bretbrownjr hinted at, sooner rather than later. Otherwise there will be a ton of copy-pasted CMake in every Beman project, and any time there's a change to the standard it will mean updating every project instead of making an update in one place. I know @bretbrownjr wasn't a fan on [this PR](https://github.com/bemanproject/exemplar/pull/115/checks) b/c of its size, but I think its somewhat representative of the amount of CMake configuration we will need. Would we really want that duplicated in every Beman repo?

I know there's concern about this adding a dependency to every Beman project, and what the mechanism will be to incorporate it into every project. I think [this discussion](https://discourse.bemanproject.org/t/make-beman-packageable/331/1) is relevant. If we agree to use `find_package(...)` and make it the library builder's responsibility to provide the package, then we can add something like `find_package(beman.config)` to every Beman project. For ergonomics, we could treat Beman configuration as a special case and support `FetchContent` if something like a `BEMAN_FETCHCONTENT_BEMAN_CONFIG` flag is set to a truthy value.

wusatosi:
@bretbrownjr @camio we need a direction forward.

Should I simply refactor version detection into the source file?

camio:
> Should I simply refactor version detection into the source file?

I don't know what you mean by this, but I [my suggestion](https://github.com/bemanproject/exemplar/pull/108/files#r1943696595) stands. The question of where to put common infrastructure is a distraction. That problem is being addressed by a separate initiative and, for now, you can just place the implementation of `beman_exemplar_check_cpp_17_support`  in the `cmake/` subdirectory.

cmake/feature_test.cmake
wusatosi: Adopted

camio (approved):
This looks good to me.

PR #113, opened 2025-02-06
Make project header only

wusatosi:
From usage experience, subsequent project usually only put source code in `include` directory and not touch the `src/` folder.

Having current `src/` folder increase setup friction, as there's an extra empty `src/beman/{NAME}/{NAME}.cpp` file that needs to be updated when adopting as subsequent library, and users doesn't realize there's another CMake file under the folder.

I don't think we should endorse header only library, but I think this is the common case.

This does conflict with #104 , @bretbrownjr

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,12 +2,7 @@
 
 cmake_minimum_required(VERSION 3.25)
 
-project(
-    beman.exemplar # CMake Project Name, which is also the name of the top-level
-    # targets (e.g., library, executable, etc.).
-    DESCRIPTION "A Beman library exemplar"
-    LANGUAGES CXX
-)
+project(beman.exemplar DESCRIPTION "A Beman library exemplar" LANGUAGES CXX)
 
 enable_testing()
 
@@ -41,7 +36,7 @@ if(BEMAN_EXEMPLAR_BUILD_TESTS)
     FetchContent_MakeAvailable(googletest)
 endif()
 
-add_subdirectory(src/beman/exemplar)
+add_subdirectory(include/beman/exemplar)
 
 if(BEMAN_EXEMPLAR_BUILD_TESTS)
     add_subdirectory(tests/beman/exemplar)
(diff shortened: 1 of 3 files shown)


neatudarius (changes requested):
@wusatosi , I think we discussed and decided not to do this. Please check sync notes.

Also, check with @bretbrownjr about https://github.com/bemanproject/beman/issues/93.


Temporary blocking this PR.

wusatosi:
@neatudarius yeah I think we discussed this before as well. But I remember the outcome to be neutral. 

I did look through the meeting minutes, this is the only piece I found to be potentially relevant to this. 

> Proposed topic: Header-only libraries deliverable? Static library vs interface (check https://github.com/beman-project/optional26/pull/75).
> Darius: Preference - interface, but add it to the Beman Standard
> ACTION: Darius, will add and entry to the standard that requires INTERFACE targets for header-only libraries.
> David: Maybe we should have a function requiring a .cpp in exemplar.

From usage experience, projects usually start off as header only and rarely use the `src` folder. It might be easier for us to grow from interface to static library instead of being a static library outright.

wusatosi:
Hey @bretbrownjr ,
Thanks for the long explanation. 

I am unfortunately (after learning so much from all the CMake related threads, maybe fortunately) not a CMake guru, and I agree with you this is very minor in comparison to other ambitions we have.

I am not arguing against header only projects.

I am not advocating to make header only library the requirement in beman. This PR is just to switch to it as default. The motivation is to reduce starting friction, as I have seen frustration from people trying to launch some new library using exemplar.

I believe moving the `src/beman/.../CMakeLists.txt` to `include/beman/.../CMakeLists.txt` and removing the need to update the no-op file at `src/beman/exemplar/identity.cpp` (thus voiding the whole `src` folder) would be a great place to start. People starts customizing exemplar by going through every file and updating every reference to exemplar with the library name of their choosing, moving the default to header only would reduce the number of files to update from 11 files to 9 + 1 folder, and move the CMake file that is under `src` to `include`, which people are more likely to be using, so it is more obvious that `include/.../CMakeLists.txt` is controlling the source set. This is on top of the observation that I think no-one puts anything in the `src` folder to start.

Well, maybe it is a sign that we desperately needs templates, and the time we spent on writing this should be on envisioning how we want to approach templating.

Let's halt this discussion for now.

PR #121, opened 2025-02-12
Add Basic CMake Config File Package Creation and Installation

chayden83:
This PR adds basic [CMake config-file package](https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#config-file-packages) creation and installation. After performing a build and install, the install destination directory now contains the following files.
```
$ find /tmp/beman/ -type f
/tmp/beman/usr/local/lib/debug/libbeman.exemplar.a
/tmp/beman/usr/local/lib/cmake/beman.exemplar/beman.exemplar-targets-debug.cmake
/tmp/beman/usr/local/lib/cmake/beman.exemplar/beman.exemplar-config.cmake
/tmp/beman/usr/local/lib/cmake/beman.exemplar/beman.exemplar-version.cmake
/tmp/beman/usr/local/lib/cmake/beman.exemplar/beman.exemplar-targets.cmake
/tmp/beman/usr/local/include/beman/exemplar/identity.hpp
```
Assuming the install location were canonical, another CMake project can import and link to the `beman.exemplar` library like so.
```cmake
find_package(beman.exemplar REQUIRED)
target_link_libraries(<my-target> [INTERFACE|PUBLIC|PRIVATE] beman::exemplar)
```
Unlike my [other PR](https://github.com/bemanproject/exemplar/pull/115), this PR is minimal and adds no support for different target export variants. The target export file is always called `beman.exemplar-targets.cmake`.

PR #122, opened 2025-02-12
Add Test for Basic CMake Config-File Package

chayden83:
This PR extends #121 to include a test to verify that the CMake config-file package is valid. That is:

- The config-file package can be used with the command `find_package(beman.exemplar REQUIRED)`.
- The config-file package correctly exports the `beman::exemplar` target.
- The exported `beman::exemplar` target can be linked.

The directory layout I chose is somewhat arbitrary, so please let me know if we should move some of the files to different directories.

Alternatively, if this is not how we want to integrate config-file package tests, please let me know.

### DIFF
--- a/src/beman/exemplar/CMakeLists.txt
+++ b/src/beman/exemplar/CMakeLists.txt
@@ -13,13 +13,36 @@ target_sources(
         FILES ${PROJECT_SOURCE_DIR}/include/beman/exemplar/identity.hpp
 )
 
-set_target_properties(beman.exemplar PROPERTIES VERIFY_INTERFACE_HEADER_SETS ON)
+set_target_properties(beman.exemplar PROPERTIES VERIFY_INTERFACE_HEADER_SETS ON EXPORT_NAME exemplar)
 
 install(
     TARGETS beman.exemplar
     EXPORT beman.exemplar
     DESTINATION
-    $<$<CONFIG:Debug>:debug/>${CMAKE_INSTALL_LIBDIR}
-    RUNTIME DESTINATION $<$<CONFIG:Debug>:debug/>${CMAKE_INSTALL_BINDIR}
+    ${CMAKE_INSTALL_LIBDIR}$<$<CONFIG:Debug>:/debug>
+    RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}$<$<CONFIG:Debug>:/debug>
     FILE_SET HEADERS DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
 )
+
+if(BEMAN_EXEMPLAR_INSTALL_CONFIG_FILE_PACKAGE)
+  include(CMakePackageConfigHelpers)
+
+  configure_package_config_file("${PROJECT_SOURCE_DIR}/cmake/${PROJECT_NAME}-config.cmake.in" "${PROJECT_BINARY_DIR}/cmake/${PROJECT_NAME}/${PROJECT_NAME}-config.cmake"
+    INSTALL_DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}"
+    PATH_VARS PROJECT_NAME PROJECT_VERSION
+  )
+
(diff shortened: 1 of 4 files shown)


src/beman/exemplar/CMakeLists.txt
```diff
@@ -13,13 +13,56 @@ target_sources(
...
+    ${CMAKE_INSTALL_LIBDIR}$<$<CONFIG:Debug>:/debug>
+    RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}$<$<CONFIG:Debug>:/debug>
     FILE_SET HEADERS DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
 )
+
+if(BEMAN_EXEMPLAR_INSTALL_CONFIG_FILE_PACKAGE)
+    include(CMakePackageConfigHelpers)
+
+    configure_package_config_file(
+        "${PROJECT_SOURCE_DIR}/cmake/${PROJECT_NAME}-config.cmake.in"
```
bretbrownjr, line 34: I'd rather not use `PROJECT_NAME` like this. It just obfuscates what's going on. The package name and the project name do match for this project, but they're different settings. I'd be fine just hardcoding `beman.exemplar` literally throughout this file, but if we want a self-documenting local variable like `export_package_name`, or something, that's fine too.

bretbrownjr (commented):
I like the scope of this PR, @chayden83. It's a big missing feature that I think we can turn this around reasonably efficiently. I like having tests in-repo as well so that mistakes are caught quickly.

As I said in other threads, I'd eventually like all of the install and export logic abstracted out to a reusable helper across Beman libraries, but it makes sense to come back to that in another PR. I do expect that C++ library engineers will consider CMake export code to be magic, understand none of it. and tolerate it at best.

Specifically on the test workflow, I'm happy to see it land now, but I will note that the whole in-repo test approach could be replaced by [what Conan calls a "test package"](https://docs.conan.io/2/tutorial/creating_packages/test_conan_packages.html) whenever we get far enough to support packaging via Conan. I expect that approach to be a lot more intuitive and maintainable. Eventually.

As to status of the PR, I have a lot of fairly modest style points I'd like you to consider. Let me know if you're OK with the suggestions and let's talk otherwise.

bretbrownjr:
Oops. I didn't realize when I opened up this PR that #122 was intended to be reviewed first. If it's too much of a pain to figure out which comments go on that PR versus this one, let me know.

If you don't mind, I'll put this PR in Draft status since it isn't ready to review until #122 is complete.

chayden83:
> Specifically on the test workflow, I'm happy to see it land now, but I will note that the whole in-repo test approach could be replaced by [what Conan calls a "test package"](https://docs.conan.io/2/tutorial/creating_packages/test_conan_packages.html) whenever we get far enough to support packaging via Conan. I expect that approach to be a lot more intuitive and maintainable. Eventually.

That makes sense. I don't have any experience w/Conan. I have done a lot of packaging, but only using Linux distro package managers.

chayden83:
Following up to see if we're ready to merge this PR. Thanks!

JeffGarland:
I merged your other PR and have updated this branch to the latest main. Looking for @bretbrownjr and others to finish any comments.

tests/beman/exemplar/CMakeLists.txt
```diff
@@ -10,3 +10,31 @@ target_link_libraries(
...
+            add_test(
+                NAME beman.exemplar.test.config-file-package
+                COMMAND
+                    "${PROJECT_SOURCE_DIR}/cmake/test/test.bat"
+                    "${CMAKE_COMMAND}"
+                    "${PROJECT_SOURCE_DIR}/cmake/test" # source dir
+                    "${CMAKE_CURRENT_BINARY_DIR}/cmake/test" # build dir
+                    "${PROJECT_BINARY_DIR}/cmake/${PROJECT_NAME}" # prefix path
+                    "$<CONFIG>"
+            )
```
ClausKlein, line 37: I don't like to call bat as sh scripts, because this is not portable.

And you see, this if else block should to the same, but it does't!

Too the `CMAKE_CXX_STANDARD` is a user defined value!

For `CXX_MODULES`, the build and test environment mast be absolute the same.

PR #128, opened 2025-02-27
bump gtest

camio:
Bump release to 1.16.0 and use GTest as the package name.
    
This should fix the issues found in https://github.com/bemanproject/exemplar/pull/92

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,14 +31,14 @@ include(GNUInstallDirs)
 if(BEMAN_EXEMPLAR_BUILD_TESTS)
     # Fetch GoogleTest
     FetchContent_Declare(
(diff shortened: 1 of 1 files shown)


wusatosi (approved):
LGTM, out of curiosity, is there a reason that we do not use `v1.16.0` as tag name?

Edit I learned something new today, [source](https://cmake.org/cmake/help/latest/module/FetchContent.html):
> Where contents are being fetched from a remote location and you do not control that server, it is advisable to use a hash for GIT_TAG rather than a branch or tag name. A commit hash is more secure and helps to confirm that the downloaded contents are what you expected.

neatudarius (commented):
LGTM

Issue #61, opened 2024-10-25
Concern about putting `FetchContent_MakeAvailable(…)` in a `block()`


dabrahams:
https://github.com/beman-project/exemplar/blob/727a5c62a60b426c859e9905c136b9012a82fa83/CMakeLists.txt#L29C1-L30C1

I don't do it for reasons explained [here](https://github.com/hylo-lang/CMakeModules/blob/6577fca51495e82a94e76a8b70113c9e6b98f539/FetchHyloDependency.cmake#L12), FWIW.

camio:
What @dabrahams linked to:

```
# Not using block() here because FetchContent_MakeAvailable typically causes dependency-specific
# global variables to be set, and I'm not sure to what extent they may be needed
```

There is no such thing as "global" variables in CMake. There are CACHE entries, which `block()` will not impact, and normal variables which have a scope that can only be propagated above one level (via `set(... PARENT_SCOPE)`). I don't understand the issue here.

dabrahams:
Fine, it should have said

```
# Not using block() here because FetchContent_MakeAvailable typically causes dependency-specific
# variables to be set in the parent scope, and I'm not sure to what extent they may be needed
```

camio:
If there are dependency-specific variables required to use a dependency, they should be CACHE variables. I can't fathom why normal variables would be used for this purpose.

dabrahams:
I can't fathom why you're talking about CACHE variables. I'm talking about variables that will be set by `FetchContent_MakeAvailable`; just look at the docs for `find_package` and note all the variables it says will be set.  Only 4 of them are documented there to be cache variables.

dabrahams:
But you know, feel free to close this issue if my concerns are misplaced.

wusatosi:
Closing this issue as it appears to be stale

camio:
I'm reopening this issue. @dabrahams is correct in that our `block()` usage can cause issues if this is used as a general pattern.

Say a library has zlib as a dependency and uses `FetchContent` to declare it. If zlib is found via `find_package` one can use the `ZLIB::ZLIB` target without hesitation. However, if the dependent wants to inquire ZLIB's version using the `ZLIB_VERSION` variable, it will not be available due to the `block()` call enclosing `FetchContent_MakeAvailable`.

wusatosi:
> I'm reopening this issue. @dabrahams is correct in that our `block()` usage can cause issues if this is used as a general pattern.
> 
[...]

PR #129, opened 2025-02-28
Use FetchContent with FIND_PACKAGE_ARGS

ClausKlein:


CMakeLists.txt
```diff
@@ -31,14 +31,21 @@ include(GNUInstallDirs)
...
     # Fetch GoogleTest
     FetchContent_Declare(
-        GTest
+        googletest
+        FETCHCONTENT_TRY_FIND_PACKAGE_MODE
+        ALWAYS
         GIT_REPOSITORY https://github.com/google/googletest.git
-        GIT_TAG
-            6910c9d9165801d8827d628cb72eb7ea9dd538c5 # release-1.16.0
+        GIT_TAG v1.16.0
```
neatudarius, line 38: I think this was already discussed and we prefer using plain commit id.

https://github.com/bemanproject/exemplar/pull/128#pullrequestreview-2649347980

Please wait for a check from @camio , as this reverts part of the diff in #128 .

 Also, I think we had the block() and we removed it in https://github.com/bemanproject/exemplar/commit/0f082a5ad5cdf796fc77481c73fb2b7383259a5b. @wusatosi , can you take a look?

camio (changes requested):
None of changes suggested here include rationale. Some of these, such as `cmake_minimum_required` and `block()/endblock()` have been discussed before with an outcome favoring the status quo.

CMakeLists.txt
wusatosi: Removing `block()` is a requested change, discussed here: #61

wusatosi: regarding the tag, use commit hash maybe preferred as best practice:

As I quoted in the [review comment](https://github.com/bemanproject/exemplar/pull/128#pullrequestreview-2649347980):
> [source](https://cmake.org/cmake/help/latest/module/FetchContent.html):
> Where contents are being fetched from a remote location and you do not control that server, it is advisable to use a hash for GIT_TAG rather than a branch or tag name. A commit hash is more secure and helps to confirm that the downloaded contents are what you expected.

I personally don't see the benefit of commit hash when the tag is specified down to patches. But I will leave the decision to @camio @bretbrownjr .

CMakeLists.txt
```diff
@@ -31,14 +31,21 @@ include(GNUInstallDirs)
...
+        FIND_PACKAGE_ARGS 1.16.0 NAMES GTest COMPONENTS gmain_main
```
wusatosi, line 39: What is this line for? @ClausKlein 

I couldn't find any corresponding documentation about this flag https://cmake.org/cmake/help/latest/command/find_package.html

bretbrownjr (changes requested):
This conflicts with the direction we polled on previously. Consensus is currently to use find_package in CMakeLists.txt and inject FetchContent when it's interesting. I have another PR open to make that change, so this PR is in conflict.

I need to update docs on the other PR, but otherwise I believe we are prepared to land it.

ClausKlein:
ok, no problem for me

CMakeLists.txt
ClausKlein: [FIND_PACKAGE_ARGS](https://cmake.org/cmake/help/latest/module/FetchContent.html#command:fetchcontent_declare)

ClausKlein:
> None of changes suggested here include rationale. Some of these, such as `cmake_minimum_required` and `block()/endblock()` have been discussed before with an outcome favoring the status quo.

You will end with a lot of cmake warnings in near future!

CMakeLists.txt
```diff
@@ -31,14 +31,21 @@ include(GNUInstallDirs)
...
+        SYSTEM
     )
-    set(INSTALL_GTEST OFF) # Disable GoogleTest installation
-    FetchContent_MakeAvailable(GTest)
+    block()
+        set(CMAKE_UNITY_BUILD ON)
+        set(BUILD_GMOCK OFF)
+        set(INSTALL_GTEST OFF) # Disable GoogleTest installation
+        FetchContent_MakeAvailable(googletest)
+    endblock()
```
ClausKlein, line 48: @camio @wusatosi 
**Note**: the `block()` scope prevent the `CMAKE_UNITY_BUILD` option to be set for your projects too!

**IMHO**: You do NOT use GMock, so you need not waste time to build it.

wusatosi: If you have strong opinion on this. You can create a PR that only contain this change with documented motivations.

PR #142, opened 2025-03-14
Set  CMAKE_PROJECT_TOP_LEVEL_INCLUDES in base preset

### LINKED ISSUE #141: exemplar build regression
On my MacOS machine, I used to be able to run `cmake --workflow --preset appleclang-release` and everything worked.

Now, I get an error

```
(base) dsankel@Davids-Mac-Studio exemplar % cmake --workflow --preset appleclang-release
Executing workflow step 1 of 3: configure preset "appleclang-release"

Preset CMake variables:

  CMAKE_BUILD_TYPE="RelWithDebInfo"
  CMAKE_CXX_STANDARD="20"
  CMAKE_TOOLCHAIN_FILE="cmake/appleclang-toolchain.cmake"

-- The CXX compiler identification is AppleClang 14.0.0.14000029
-- Check for working CXX compiler: /usr/bin/clang++ - skipped
-- Found GTest: /Users/dsankel/miniconda3/lib/cmake/GTest/GTestConfig.cmake (found version "1.10.0")
-- Looking for __cpp_lib_ranges - not found
CMake Warning at examples/CMakeLists.txt:12 (message):
  Missing range support! Skip: identity_as_default_projection


Examples to be built: identity_direct_usage
-- Configuring done (2.2s)
-- Generating done (0.0s)
-- Build files have been written to: /Users/dsankel/repos/bemanproject/exemplar/build/appleclang-release

Executing workflow step 2 of 3: build preset "appleclang-release"

[2/6] Linking CXX static library src/beman/exemplar/libbeman.exemplar.a
warning: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: archive library: src/beman/exemplar/libbeman.exemplar.a the table of contents is empty (no object file members in the library define global symbols)
[6/6] Linking CXX executable tests/beman/exemplar/beman.exemplar.tests.identity
FAILED: tests/beman/exemplar/beman.exemplar.tests.identity tests/beman/exemplar/beman.exemplar.tests.identity[1]_tests.cmake /Users/dsankel/repos/bemanproject/exemplar/build/appleclang-release/tests/beman/exemplar/beman.exemplar.tests.identity[1]_tests.cmake 
: && /usr/bin/clang++ -O3  -O2 -g -DNDEBUG -arch arm64 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.1.sdk -Wl,-search_paths_first -Wl,-headerpad_max_install_names  tests/beman/exemplar/CMakeFiles/beman.exemplar.tests.identity.dir/identity.test.cpp.o -o tests/beman/exemplar/beman.exemplar.tests.identity  src/beman/exemplar/libbeman.exemplar.a  /Users/dsankel/miniconda3/lib/libgtest_main.dylib  /Users/dsankel/miniconda3/lib/libgtest.dylib && cd /Users/dsankel/repos/bemanproject/exemplar/build/appleclang-release/tests/beman/exemplar && /opt/homebrew/Cellar/cmake/3.30.1/bin/cmake -D TEST_TARGET=beman.exemplar.tests.identity -D TEST_EXECUTABLE=/Users/dsankel/repos/bemanproject/exemplar/build/appleclang-release/tests/beman/exemplar/beman.exemplar.tests.identity -D TEST_EXECUTOR= -D TEST_WORKING_DIR=/Users/dsankel/repos/bemanproject/exemplar/build/appleclang-release/tests/beman/exemplar -D TEST_EXTRA_ARGS= -D TEST_PROPERTIES= -D TEST_PREFIX= -D TEST_SUFFIX= -D TEST_FILTER= -D NO_PRETTY_TYPES=FALSE -D NO_PRETTY_VALUES=FALSE -D TEST_LIST=beman.exemplar.tests.identity_TESTS -D CTEST_FILE=/Users/dsankel/repos/bemanproject/exemplar/build/appleclang-release/tests/beman/exemplar/beman.exemplar.tests.identity[1]_tests.cmake -D TEST_DISCOVERY_TIMEOUT=5 -D TEST_XML_OUTPUT_DIR= -P /opt/homebrew/Cellar/cmake/3.30.1/share/cmake/Modules/GoogleTestAddTests.cmake
dyld[77184]: Library not loaded: @rpath/libgtest_main.dylib
  Referenced from: <84E54742-FDDC-3A84-A810-19B850FA2CAD> /Users/dsankel/repos/bemanproject/exemplar/build/appleclang-release/tests/beman/exemplar/beman.exemplar.tests.identity
  Reason: tried: '/usr/local/lib/libgtest_main.dylib' (no such file), '/usr/lib/libgtest_main.dylib' (no such file, not in dyld cache)
CMake Error at /opt/homebrew/Cellar/cmake/3.30.1/share/cmake/Modules/GoogleTestAddTests.cmake:112 (message):
  Error running test executable.

    Path: '/Users/dsankel/repos/bemanproject/exemplar/build/appleclang-release/tests/beman/exemplar/beman.exemplar.tests.identity'
    Working directory: '/Users/dsankel/repos/bemanproject/exemplar/build/appleclang-release/tests/beman/exemplar'
    Result: Subprocess aborted
    Output:
      

Call Stack (most recent call first):
  /opt/homebrew/Cellar/cmake/3.30.1/share/cmake/Modules/GoogleTestAddTests.cmake:226 (gtest_discover_tests_impl)


```

My expectation was that the presets would still use FetchContent for dependencies because user environments are typically weird and frequently broken.

steve-downey:
Set  CMAKE_PROJECT_TOP_LEVEL_INCLUDES to our GitHub dependency provider as a common default so that common out of the box workflows will contintue to work.

Edit from river: fix #141


wusatosi:
I have entertained this idea with @bretbrownjr at #105 (the PR that introduced this fetch content delegation).
You can read our discussion here: https://github.com/bemanproject/exemplar/pull/105#issuecomment-2692451900 .

I think this is a good short-term patch, but I think we might need to essentially redesign presets to fit this combination explosion. 

I am in support of this change, but I will leave the decision to @bretbrownjr and @camio .

steve-downey:
I saw @camio commenting earlier that we broke his workflow, and I think this minimal solution is OK for out of the box use for someone kicking the tires, or in one of our common cases. 

Presets just aren't designed to support all the possible combinations, though. They're for a small-ish set for a group who builds things in a very consistent way, sharing the same workflow, with a small set of tools involved. 

That is a very common situation.  But it's really not for the matrix explosion of a public C++ project where you want to check all the things all the ways.

steve-downey:
We might be able to use `condition:` blocks and `$env{<variable-name>}` to provide some ways of injecting options into the presets files, but the fact that it's not obvious how to do this suggests that it's not going to be well supported? I've never used presets in anger, and we've not adopted them at Bloomberg, sticking with our own workflow tools and configuration management strategies, which are unfortunately far too tied to our environment to be usefully released.

bretbrownjr:
Except FetchContent out of the box also breaks other workflows. I started this packaging effort a few months ago because my out of the box experience wasn't working.

I'm not opposed to some branded workflows obviously supporting FetchContent. I don't agree that it's a (singular) sensible default though.

bretbrownjr (changes requested):
I'd like to discuss this more.

I am not opposed to presets in this direction. I'm not convinced it goes in the root preset however.

bretbrownjr:
The reasons Bloomberg has been slow to adopt presets have more to do with the design limitations of presets than any in-house tech conflicts.

Namely:

* Unbounded complexity when configuring all interesting build flavors 

* The need to maintain preset files in every. single. repo. The tech is nowhere near settled enough to land one and never touch it for five plus years. Clang is still coming out with new sanitizers.

* The biggest headaches have to do with dependency management, not CMake build workflow defaults. And good solutions to dependency management alleviate a lot of the pain points that presets help with (like better defaults).

camio:
I'm supportive of this PR. There are tradeoffs for whatever default we choose, but a FetchContent default has a fully known dependency set which makes issues at first-exposure (like #141) much less likely.

JeffGarland:
I've also approved for at least the short term due to the regression.  @bretbrownjr are we ok to fix and open another issue to re-evaluate?

JeffGarland (approved):

neatudarius (approved):
LGTM 👍 

I'm also OK to merge this right now and do follow-up PRs.

bretbrownjr (approved):
Moving my "changes requested" to "approval" to unblock someone merging. I still don't think setting all presets to FetchContent workflow is a good idea, however.

neatudarius:
Note: I fixed the conflicts and merged with main. If all tests will pass, I will merge this PR as now it has strong consensus.

PR #145, opened 2025-03-16
CMake workflows build in parallel by default

ednolan:
Setting `"jobs": 0` will make CMake use the native build tool's default number of concurrent processes, which is equal to `$(nproc)` for ninja.


wusatosi (approved):
I haven't found setting jobs to 0 paralleling the build process to be a documented feature. But this is working. 

Confirm applied to all build presets.

PR #150, opened 2025-03-30
Reuse infra

Jason5480:
## Description

This MR's goal is to trigger a discussion on the role of the infra repo and the relation with the exemplar. It proves that specific folders and files from the infra repo can be synchronized.

## Motivation and Context

All the libraries that used exemplar as a starting template have some infrastructure code that needs to be upgraded after the creation of the library. This is challenging to do without interrupting the development lifecycle and breaking things that worked before. On the other hand, if you let the library developer update the infrastructure manually, he most likely won't risk it, leaving the repo container ci pipelines in an old state. Since all of the Beman repos are C++ libraries with the same needs, it should be one place where those changes are made the infra repo, and the update should be done either manually by running the pull_infra.sh script or automatically after a successful release.

## Testing

Having installed the prerequisites described in the pull_infra.sh go to the root of the repo and run
`./pull_infra.sh -d "devcontainer .github" -f "README.md"' AND `.\pull_infra.py -d devcontainer`  (or any folder or file that exists in infra repo)
The script using the verdor workflow, should add/update the infra remote and put it in a temporary folder using the git subtree.
Then it should sync the folders and files that passed as arguments, commit the changes and clean up the residuals.

## Meta


.gitattributes
```diff
@@ -0,0 +1,8 @@
+###############################
```
neatudarius, line 1: Missing SPDX identifier in both files.

neatudarius:
Thanks for initiative! I created an issue for you - https://github.com/bemanproject/infra/issues/6

.gitattributes
Jason5480: I added the SPDX line in both. Shall we do this for every file? For example, gitignore also doesn't have this.

Jason5480:
> Thanks for initiative! I created an issue for you - [bemanproject/infra#6](https://github.com/bemanproject/infra/issues/6)

After verifying the correctness of the script, it would be useful to make a table of the infrastructure files/folders that we want to get updated after the creation of the library (from the exemplar template). For now, I see the .devcontainer and .github folders as main candidates so we can make changes in the CI pipelines and container and mass update the changes in the library repos.

neatudarius:
For a long term solution, probably some of the files will be sync with such script, some of them will be shipped as git actions (e.g. linters).

So for now I would keep it simple: just a table which sets a tag (deep copy, git action etc).

pull_infra.sh
```diff
@@ -0,0 +1,159 @@
+#!/bin/bash
```
camio, line 1: Would it make sense to make this into a cmake script so we get Windows support?

Jason5480: I would also like this functionality to be cross-platform. However, I think that CMake is not the right option here, since the sync should be unrelated to the configure-build-test steps. I will try to push a python version of it.

Jason5480: With a little help from AI I created a python version and I run it successfully on windows with python 3.12 installed.

pull_infra.py
```diff
@@ -0,0 +1,137 @@
...
+import sys
+from pathlib import Path
+
+
+def run(cmd, cwd=None, check=True):
+    print(f"🔧 Running: {' '.join(cmd)}")
+    result = subprocess.run(cmd, cwd=cwd, capture_output=True, text=True)
+    if check and result.returncode != 0:
+        print(f"❌ Error: {result.stderr.strip()}")
+        sys.exit(result.returncode)
```
bretbrownjr, line 16:
```suggestion
    result = subprocess.run(cmd, cwd=cwd, capture_output=True, check=check, text=True)
```

`subprocess.run` has a `check` feature that can save some complexity here. If you don't want to use it for some reason, I'd appreciate a comment to explain why not.

pull_infra.sh
bretbrownjr: You can write a full program in CMake and run it with `cmake -P arbitrary_file_name`. See the docs or `--help` for the `cmake` command for reference.

If you name the script `pull_infra`, make it executable, and add `#!/usr/bin/env cmake -P` as the shebang line, the user would just run it and not care what language it's written in. It would also be awkward to intentionally pull into a `CMakeLists.txt` on accident because the lack of a standard name for a CMake file.

bretbrownjr: That being said, given we seem fine using precommit for SDLC purposes, I don't mind shipping other utilities in python. But it's true that a fairly simple workflow like this could be written as a standalone script in CMake with fewer dependencies for the users we're targeting. It would be harder to maintain, though, probably.

pull_infra.py
```diff
@@ -0,0 +1,137 @@
...
+    return result.stdout.strip()
+
+
+def vendor_branch_exists(branch_name):
+    return branch_name in run(["git", "branch"], check=False)
```
bretbrownjr, line 21:
```suggestion
    return branch_name in run(["git", "branch", "--all"], check=False)
```

This is an actual bug because the function doesn't do what it says it does.

If the goal of the function is to determine if the *current* branch is the given branch, the current implementation works, but we need a new name for the function.

See:
https://git-scm.com/docs/git-branch

pull_infra.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+def vendor_remote_exists(remote_name):
+    return remote_name in run(["git", "remote"], check=False)
```
bretbrownjr, line 25:
```suggestion
    return remote_name in run(["git", "remote", "--verbose"], check=False)
```

Similar problem as for the `vendor_branch_exists` function, though it's git, so the flags aren't spelled consistently 🤷‍♂️ .

See:
https://git-scm.com/docs/git-remote

pull_infra.py
```diff
@@ -0,0 +1,137 @@
...
+
+        if local_hash == remote_hash:
+            print("✅ Vendor repository is already up to date. No new changes found.")
+            vendor_cleanup(vendor_prefix, vendor_local_branch, vendor_name)
+            return
+
+        # Checkout or create vendor tracking branch
+        if vendor_branch_exists(vendor_local_branch):
+            run(["git", "checkout", vendor_local_branch])
+            run(["git", "pull", vendor_name, vendor_remote_branch])
```
bretbrownjr, line 91: This should probably be a less expensive and destructive `git merge --ff-only`. `pull` does off-box I/O. I would expect that we're already synced up from the above `git fetch`.

pull_infra.py
```diff
@@ -0,0 +1,137 @@
...
+        remote_hash = run(["git", "rev-parse", f"{vendor_name}/{vendor_remote_branch}"])
+
+        if local_hash == remote_hash:
+            print("✅ Vendor repository is already up to date. No new changes found.")
+            vendor_cleanup(vendor_prefix, vendor_local_branch, vendor_name)
+            return
+
+        # Checkout or create vendor tracking branch
+        if vendor_branch_exists(vendor_local_branch):
+            run(["git", "checkout", vendor_local_branch])
```
bretbrownjr, line 90: Modern practice is to use `git switch` instead of `git checkout` when appropriate. It has fewer footguns when all you want to do is switch branches, like in this script.

bretbrownjr (changes requested):
I have some requests for the python script before going with that.

We should also get consensus from contributors about how they get these scripts. Options include at least:

* Maintain a copy of a script like this per project
* `pip install` it (perhaps in a venv and/or via precommit) from GitHub.com
* `pip install` it from PYPI (more infra work for this of course)
* Just download and run the script as needed via `curl ...` and then run it

bretbrownjr:
But thanks for taking a look at this problem. It's an important one and we do need ideas and contributions on tackling it!

bretbrownjr:
Also poke @steve-downey as he has similar opinions about using `git subtree` for vendoring workflows.

steve-downey:
Will catch up when I'm not traveling. The real drawback with subtree is leaving a signpost that the git objects come from, or are shared by, a different repo so that it's obvious how to merge new changes in.
The advantage is that any changes are an actual commit in the repo, so everyone gets the same thing on a checkout.

steve-downey:
I definitely like the overall approach here, and it's very close to what I do manually. I tend to _not_ move files out of the vendored location in order to facilitate merging changes back to the shared source for infrastructure. Of course I'm basically modelling the `git submodule` workflow only with better integration with the rest of my git process. I don't have something I can share, unfortunately, as I've been doing it for $work related projects. 

But particularly for infrastructure, changes aren't going to start in the `infra` repository, they're going to be created, tested, and modified in actual work in anger, and then merged back up for use by other projects. At which point we'll discover something that needs more generalization. The nice thing about a subtree is that you can effectively carry a patch in projectA and still merge changes from upstream, even if you can't ff-only the commits. 

If we're going to move to python, and I am fine with that as a baseline dependency, we should drop the shell script version so we don't multiply the maintenance burden.  This is on the edge the size of what I'd like to see in a shell script, anyway. 

I wouldn't want to try to have this automated in an actual $GIT_DIR/hooks/ script. Unless the success rate is in the 5-9s range, breaking a project automatically is a big risk. I haven't tried writing my own [precommit ](https://pre-commit.com/) hooks yet, so I'm not sure how that would work. But I think a github action that we could share from the infra repository (?) that could be run on a work repository to check that CI otherwise still works might be a place to start? In any case, the place to start isn't automating the process, it's working out what an automatable process might be and executing that by hand.

steve-downey:
BTW, for anyone trying to work with whole org projects, I have a few scripts at https://github.com/steve-downey/mirror-beman-scripts to help get a mirror into our internal github enterprise  server. No promises or warranty, but https://github.com/steve-downey/mirror-beman-scripts/blob/main/clone-beman.sh might save someone a few minutes in figuring out how to clone all of https://github.com/bemanproject/

bretbrownjr (changes requested):
It sounds like there's reasonable support for a script like this (so far).

That being said, I doubt we want both bash and python versions of the `pull_infra` script.

pull_infra.py
Jason5480: It is probably better that way.

pull_infra.sh
Jason5480: I didn't know that I could run it with cmake like that. Anyways, I think that python is fine for now and it is not an overhead, since we already installed it in our environment due to precommit and also other C++ tooling like conan uses python.

pull_infra.py
Jason5480: verbose will add the url in the list output and I just need the remote name

pull_infra.py
Jason5480: At this stage we check if the vendor branch already exists locally (probably from a previous unsuccessful run) if this is the case, we pull the latest changes. Otherwise, we checkout and track the remote branch.

Jason5480:
> I have some requests for the python script before going with that.
> 
> We should also get consensus from contributors about how they get these scripts. Options include at least:
> 
> * Maintain a copy of a script like this per project
> * `pip install` it (perhaps in a venv and/or via precommit) from GitHub.com
> * `pip install` it from PYPI (more infra work for this of course)
> * Just download and run the script as needed via `curl ...` and then run it

I would suggest keeping it simple for now. Let's have a copy of it in the exemplar repo.
All repos from now on (after the merge of this PR) that start from the exemplar they will have it;
I also have to put it in the library repos that exist already, if this happens only once; it is fine by me.
However, this assumes that the pull_infra.py script will not need any update on its source code. If this is not the case, I am in favor of putting the pull_infra.py to the infra repo and downloading the latest version.
But then if you have a mechanism to download any file why not also download the .devcontainer and .github folders the same way?

steve-downey:
> > I have some requests for the python script before going with that.
> > We should also get consensus from contributors about how they get these scripts. Options include at least:
> > 
> > * Maintain a copy of a script like this per project
> > * `pip install` it (perhaps in a venv and/or via precommit) from GitHub.com
> > * `pip install` it from PYPI (more infra work for this of course)
> > * Just download and run the script as needed via `curl ...` and then run it
> 
> I would suggest keeping it simple for now. Let's have a copy of it in the exemplar repo. All repos from now on (after the merge of this PR) that start from the exemplar they will have it; I also have to put it in the library repos that exist already, if this happens only once; it is fine by me. However, this assumes that the pull_infra.py script will not need any update on its source code. If this is not the case, I am in favor of putting the pull_infra.py to the infra repo and downloading the latest version. But then if you have a mechanism to download any file why not also download the .devcontainer and .github folders the same way?

Similar to how rustup, ghcup, pyenv, oh-my-zsh, etc have a way of updating themselves so they can do new things for the components they manage. 

At some point the core script might need to learn to stop after updating itself and letting the maintainer know to retry. Some extra steps should be OK as this isn't a normal end-user operation, any more than updating googletest to a new release is.

ednolan:
Maybe I'm missing something but I don't understand the motivation for this script. Why wouldn't we just add the `infra` repo to `exemplar` as a git submodule?

Jason5480:
> I would prefer to add the `infra` repo to `exemplar` as a git submodule, which would remove the need for this script.

We considered this option since git submodules are most widely used. There are some tradeoffs between submodules and subtree. Git subtree was initially suggested by @steve-downey on discourse https://discourse.bemanproject.org/t/should-we-have-a-dedicated-repo-for-infrastructure/226/9.
After I searched about pros and cons of the two approaches, I came up with a rule of thumb that if the files you want to pull in are strictly separated (like a library) use submodules, but if they are part of your project (like devcontainer and github piplines) use subtree.
Also, subtree has the benefit that it is transparent to the user and the CI, since no extra git commands are needed to initialize the repository, and as an extent it adds no extra complexity. Of course, git subtree can be used without this script, but since it is less popular, I created this script as a PoC and an easy way to pull w/e files/folders are needed.

wusatosi:
Is this proposal in conflict with bx? @bretbrownjr

camio:
@ednolan, where are we with this PR? Your work at C++Now last week seems related.

ednolan:
[My script](https://github.com/bemanproject/infra/blob/2164f14d1f8e35644f14678f5ee264155f57209b/bemanmodule.py) is an alternative approach to this script, which is very similar, except that it doesn't use git subtree commands, it always pulls in the entire sub-repo instead of picking out files and folders from it, and it puts information about the sub-repo in special dotfiles rather than hardcoding it into the script. My script would also live in infra itself rather than adding it to exemplar as this PR does; I'm planning on adding some functionality that makes bootstrapping it into a Beman repository easy for maintainers. My script isn't ready to be merged yet; I have to implement some more features, bugfixes, tests, and documentation, which I'm planning to do over the weekend.

Jason5480:
The changes requested by @bretbrownjr have been made. We need an easy way to update our infrastructure for our repos, now if we use subtree or submodule, it is an implementation detail to the user imho. I went with subtree because @steve-downey suggested it and it also I saw it had some benefits (it doesn't download all the repos history). The difficult part is to bootstrap this into existing repos and also make it update itself. I went with the simplest approach to have this script in every beman library repo. I would like to see @ednolan approach to this.

camio:
A C++Now, the consensus was to go ahead with the "beman module" approach instead of git subtrees. I think once those PRs merge, we can close this one. @Jason5480, thanks for creating this PR to spawn all the discussion! Is there anything here that would be useful to port over to the other PRs?

PR #154, opened 2025-04-22
Clean up install logic

bretbrownjr:
## Description

Cleaned up and fixed installation logic.

## Motivation and Context

Using the parameters to describe the project in `project()` for installation logic is misleading. It happens to be that the name `beman.exemplar` is used as the name of the project and as the CMake export name, but those are two different concepts that do not necessarily align.

Also, the variable dereferencing in CMakeLists.txt is no less error prone than typing beman.exemplar. That's because it is not difficult to type things like ${PROJET_SOURCE_DIR}, which CMake will expand to an empty string.

Instead, the readability of the install logic improves because it looks like a file with a specific name is being specifically created and referenced, which is the intention of the relevant installation logic.

## Summary of Changes

* Remove references to PROJECT_NAME, PROJECT_SOURCE_DIR, and PROJECT_BINARY_DIR. Prefer relative paths to CMAKE_CURRENT_SOUCE_DIR and CMAKE_CURRENT_BINARY_DIR instead.
* Rename the install COMPONENT for beman.exemplar to "beman.exemplar" instead of "development".
* Add the install COMPONENT to the installation of the beman.exemplar library target itself.
* Remove extraneous FetchContent include.

## Testing

Existing CI checks should confirm no regressions. Additional manual installation testing was performed, using `cmake --install` with combinations of `--component beman.exemplar`, `--config Debug`, `--config Release`, etc.

## Meta


neatudarius (approved):
LGTM 👍 , but I'm not a CMake expert. :D 

@steve-downey @camio , I know you had some feedback about our CMake files at some point. Any feedback here?

src/beman/exemplar/CMakeLists.txt
```diff
@@ -9,8 +11,9 @@ target_sources(
     beman.exemplar
     PUBLIC
         FILE_SET HEADERS
-        BASE_DIRS ${PROJECT_SOURCE_DIR}/include
-        FILES ${PROJECT_SOURCE_DIR}/include/beman/exemplar/identity.hpp
+        BASE_DIRS ${CMAKE_CURRENT_SOURCE_DIR}/../../../include
+        FILES
+            ${CMAKE_CURRENT_SOURCE_DIR}/../../../include/beman/exemplar/identity.hpp
```
steve-downey, line 16: I dislike the connection at a distance this creates, but I don't have a really good solution without a lot more of reorganization. 

For optional the header file set is set up in the `include/beman/optional` directory. 

https://github.com/steve-downey/optional/blob/960da87ae54d817bc6f46fe9fda776b21f88e939/include/beman/optional/CMakeLists.txt#L7

Putting on my list to discuss over lunch in Aspen.

src/beman/exemplar/CMakeLists.txt
```diff
@@ -31,31 +34,31 @@ if(BEMAN_EXEMPLAR_INSTALL_CONFIG_FILE_PACKAGE)
     include(CMakePackageConfigHelpers)
 
     configure_package_config_file(
-        "${PROJECT_SOURCE_DIR}/cmake/${PROJECT_NAME}-config.cmake.in"
-        "${PROJECT_BINARY_DIR}/cmake/${PROJECT_NAME}/${PROJECT_NAME}-config.cmake"
-        INSTALL_DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}"
+        "${CMAKE_CURRENT_SOURCE_DIR}/beman.exemplar-config.cmake.in"
```
steve-downey, line 37: I'm in favor of use before reuse, so I don't have reason not to do this. 
Is there a technical reason not to use the PROJECT_NAME defined in https://github.com/bemanproject/exemplar/blob/fb698c289ea060c2ef64b98486dd2cb1618596b7/CMakeLists.txt#L6
at the top-level CML.txt? 
It does seem a little fragile, and might not compose well?

src/beman/exemplar/CMakeLists.txt
```diff
@@ -31,31 +34,31 @@ if(BEMAN_EXEMPLAR_INSTALL_CONFIG_FILE_PACKAGE)
...
     install(
         FILES
-            "${PROJECT_BINARY_DIR}/cmake/${PROJECT_NAME}/${PROJECT_NAME}-config.cmake"
-            "${PROJECT_BINARY_DIR}/cmake/${PROJECT_NAME}/${PROJECT_NAME}-version.cmake"
-        DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}"
-        COMPONENT development
+            "${CMAKE_CURRENT_BINARY_DIR}/beman.exemplar-config.cmake"
+            "${CMAKE_CURRENT_BINARY_DIR}/beman.exemplar-version.cmake"
+        DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/beman.exemplar"
+        COMPONENT beman.exemplar
```
steve-downey, line 54: > Specify an installation component name with which the install rule is associated, such as `Runtime` or `Development`. 
https://cmake.org/cmake/help/latest/command/install.html

Is there a reason to not use a generic name for the component?

steve-downey: Discussed off-line. This works around some peculiar cases with composition of cmake projects, and in particular the fragility of PROJECT_NAME and how it is set and propoagated.

steve-downey (approved):
LGTM to merge. 
We can continue to discuss improvements.

PR #158, opened 2025-04-29
Add beman.infra as a git submodule and get toolchain files from it

ednolan:
Previously, these toolchain files were part of beman.exemplar, but by splitting them out into a separate repository, we allow them to be more easily reused by other Beman repositories, helping us to distribute changes to the toolchain files and partly mitigating bitrot of the CMake files after a developer copies them from the exemplar template.


ednolan:
Abandoned in favor of using beman_module.py.

PR #160, opened 2025-04-29
Revert complications voided by #142

wusatosi:
## Description

Revert CI changes introduced in #105, this is possible because we have fix preset regression in #105 .

## Related Issues

See also: #141

## Motivation and Context

Simplify CI.

## Testing

CI passing.

## Meta

### DIFF
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -47,11 +47,7 @@ jobs:
         with:
           arch: x64
       - name: Run preset
-        run: |
-          cmake -DCMAKE_PROJECT_TOP_LEVEL_INCLUDES="./cmake/use-fetch-content.cmake" --preset ${{ matrix.presets.preset }}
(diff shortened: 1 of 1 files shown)


camio (approved):

neatudarius (approved):

.github/workflows/ci_tests.yml
```diff
@@ -238,7 +234,6 @@ jobs:
         run: |
           sudo add-apt-repository universe
           sudo apt-get update
-          sudo apt-get install -y libgtest-dev
```
ednolan, line 241: This seems to be a drive-by fix related to the fact that the ci images now contain libgtest-dev; this is a nitpick, but I think this line should have been in a separate pull request.

ednolan: Oh okay, I didn't understand that. Disregard my comment, in that case

PR #165, opened 2025-05-25
Config folder

wusatosi:
## Description

Move linter files under .config folder to avoid clutter.

## Motivation and Context

There is an increasing amount of linters installed, each with a potential config files, it make sense for us to centralize linter configs under the same folder, especially we can specify the config path easily in precommit config. With this PR, the only linter related config in the root directly is the pre-commit config.

This is also easier to symlink once we figure out how we want to distribute these files.

## Testing

CI test on `pre-commit`, I have manually disrupted all the config files and made sure pre-commit fail due to not being able to parse the config file.

## Meta


wusatosi:
@bretbrownjr should we move `lock.json` under `.config` as well?

steve-downey:
The problem with moving the config files is convincing every other tool in existence to find and apply that config rather than a default one.
Root of the project directory may not be the best place, but it's the accepted consensus.

wusatosi:
> The problem with moving the config files is convincing every other tool in existence to find and apply that config rather than a default one. Root of the project directory may not be the best place, but it's the accepted consensus.

The nice thing is all linters usually have an command line argument to specify the config file. Since we are invoking them via precommit, it's really easy to config this and gives no cost on the user the side

neatudarius:
@wusatosi , don't we need to also have the pre-commit config in .config? I suppose, next step is to retrieve `.config/` directory from `infra/` repo, right?

wusatosi:
>  Don't we need to also have the pre-commit config in .config?

No we still need pre-commit to be in the root folder, otherwise we have to specify pre-commit config location on every single execution.

The reason we can do this to other configs is because the linters are invoked by pre-commit and we can config this easily without any complication on the user end. 

>  I suppose, next step is to retrieve `.config/` directory from `infra/` repo, right?

We can evaluate this later, but definitely a possibility.

camio:
> The problem with moving the config files is convincing every other tool in existence to find and apply that config rather than a default one.
> Root of the project directory may not be the best place, but it's the accepted consensus.

I agree with Steve here. The clang-format command-line tool which is used by many editor integrations, for example, looks for a .clang-format file in one of the parent directories of the file being formatted. Putting these files in a `.config` subdirectory is not the industry practice and breaks things.

If folks disagree, I'd be happy to have a further discussion on discourse to see if there's a consensus.

Issue #166, opened 2025-05-27
CTest Discussion Redux


nickelpro:
I would like to talk about enabling testing in Beman repos

## History

This discussion now has a bit of a history:

* The introduction of unconditional `include(CTest)`: https://github.com/bemanproject/exemplar/pull/91
* To `enable_testing()` + `CTestUseLaunchers`: https://github.com/bemanproject/exemplar/pull/109
* Back to `include(CTest)`: https://github.com/bemanproject/exemplar/pull/153

And finally I stumbled into the issue updating `beman.scope`: [Update CMakeLists, prepare for packaging #31](https://github.com/bemanproject/scope/pull/31/files#r2109820703)

I think the above fully hash out the issue, and this now comes down to a direct discussion between the handful of us who care way too much about the intersection of CMake and CTest.

## Discussion

For the reasons discussed in @purpleKarrot's PR and my comment on `beman.scope`, unconditional test targets via `include(CTest)` are not workable in the general case. They do not exist on all generators, and they can be misleading even in cases where they do exist.

The generalize-able and fully understood behavior is that if you try to run tests and the tests have not been built, the command should fail by one mechanism or another. The intent of the user is to run the tests provided by the project, and if they didn't build the tests they shouldn't be told a test run passed. Moreover, the mechanism for running tests is `ctest`, the target driven approach is entirely superfluous. 

The only remaining question in my mind is if the cornucopia of other targets dragged in by `include(CTest)` is desirable for Beman. My answer to this is No. At best they annoyingly overflow target lists, at worst they borderline violate **[CMAKE.PASSIVE_PROJECTS]**.

## Proposal

I want to switch Exemplar and the other Beman projects back to the approach used in  https://github.com/bemanproject/exemplar/pull/109. I'm uncertain if this decision is worthy of the Beman standard. I would want it as something even less than a recommendation, just, "this is how we do it in Beman because a decision needed to be made".

I do not think the other approaches are invalid always, or even in bad taste, but I don't think they meet Beman's particular use case.

wusatosi:
Also CC @purpleKarrot 

> I recommend a foot race across Central Park at 72nd St

You gotta buy my ticket to fly to NYC.

purpleKarrot:
+1 to everything @nickelpro wrote. Except:

> If ..., **then** 'make test' works everywhere.

Not even then. Note that 'test' is not one of the predefined targets for autotools: https://www.gnu.org/software/automake/manual/html_node/Standard-Targets.html

[...]

PR #159, opened 2025-04-29
Refactor ci

wusatosi:
## Description

Simplify main build-and-test CI pipeline by extracting CMake invocation as a separate module.

## Motivation and Context

This is to prepare for the infra repo, we can ship in this CMake invocation as a standalone GitHub action package or using symlink with git submodule (see: #158 )

## Testing

CI passing.

## Meta

- [ ] If all approvals are obtained and the PR is green, any Beman member can merge the PR.

PR #167, opened 2025-05-31
Update Exemplar build to be modules-first

nickelpro:
This is a big block of changes (well, as big as it gets with exemplar) that does two things:
* Incorporates lessons learned from updating `beman.scope` to be packageable, as well as discussions had at C++Now
* Converts `beman.exemplar` to be "modules-first", which is to say it requires module support from the toolchain in order to run tests and build examples, but still supports using the traditional header files when installed as a library

This does not violate anything in the Beman standard, but rather makes use of things we might want to have concrete consensus on now that they've arisen as issues.

I do not intend for this to be merged outright, it's more a piece of performance art. It's easier to debate pros and cons with a concrete implementation of all those pros and cons laid out.

**Related Discussions:**
* All of the commentary from https://github.com/bemanproject/scope/pull/31 is the same here, please give that a read through.
* This re-asserts the changes being discussed in https://github.com/bemanproject/exemplar/issues/166, so please divert testing discussion there
* For some background on why I did this, see the modules discussion here https://github.com/bemanproject/scope/pull/34, but this is an appropriate PR to kvetch about how hard modules are if you need to get that out of your system. I am a sympathetic shoulder.


nickelpro:
Well the good news is the lint passes.

Most of the test failures are on compilers that don't support modules and the MSVC ones are... something

tests/beman/exemplar/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ find_package(GTest REQUIRED)
 
 add_executable(beman.exemplar.tests.identity)
 target_sources(beman.exemplar.tests.identity PRIVATE identity.test.cpp)
+target_compile_features(beman.exemplar.tests.identity PRIVATE cxx_std_20)
```
JeffGarland, line 7: what is the magic with cxx_std_20?  I mean isn't that already assured if the CMAKE_CXX_VERSION is set to 20 or above?

nickelpro: So it conflicts with **[CMAKE.PASSIVE_TARGETS]** but there's not really a way around it.

**[CMAKE.PASSIVE_TARGETS]** is about _source code_ compatibility. We shouldn't tell users building a Beman project what version they want to build with, if the source code does or does not compile is something they'll discover pretty easily.

The problem is that we need _build_ compatibility. In order to generate the packaging files correctly, CMake needs a module scanner. CMake runs the module scanner on targets and source files on C++ files it knows support at least C++20, from the docs:

> Note that scanning is only performed if C++20 or higher is enabled for the target and the source uses the CXX language. Scanning for modules in sources belonging to file sets of type `CXX_MODULES` is always performed.

So even if the source code supports something less than C++20, and you want to run your overall build as something lower than C++20 (using the header files instead of `import beman.exemplar`), for this little thing I need to turn on at least C++20 because otherwise the packaging explodes when CMake tries to figure out what to do with the module files.

nickelpro: Actually I might not need this, let me check

CMakeLists.txt
```diff
@@ -30,11 +29,80 @@ option(
     ${PROJECT_IS_TOP_LEVEL}
 )
 
-include(CTest)
+add_library(beman.exemplar)
```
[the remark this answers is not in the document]
nickelpro: Modules are nominally orthogonal to static vs dynamic libraries, interface units describe both.

So I leave it up to the person doing the build. The `BUILD_SHARED_LIBS` variable controls what `add_library` does by default. It is a question of style whether to define independent targets for static vs dynamic builds, some people swear by it, others hate it. Exemplar was already doing this so I left it as is.

JeffGarland:
> So the CI has been lying a little bit for a long time. If you wanted to test your code on GCC 13 and Clang 18 though it worked _really_ good. Very tested.

Good catch -- obviously no one is looking closely at the CI lol

wusatosi:
CI fail is not only because proper version is not selected.

Please check compiler test seris, e.g. llvm 18 CI test.
Error message is as following:
```
/home/runner/work/exemplar/exemplar/examples/identity_direct_usage.cpp:6:8: fatal error: module 'beman.exemplar' not found
    6 | import beman.exemplar;
      | ~~~~~~~^~~~~
```
See: https://github.com/bemanproject/exemplar/actions/runs/15358684463/job/43222580631?pr=167#step:5:69

tests/beman/exemplar/CMakeLists.txt
wusatosi, line 7: This violates [passive project requirement](https://github.com/bemanproject/beman/blob/5c0b8662f9e6ee7b995d5e14f4c690a6aac5ce7a/docs/BEMAN_STANDARD.md?plain=1#L344-L356) btw.

> `target_compile_features` (e.g., `cxx_std_20`) must not be used
because it modifies the compilation environment of dependent targets.

src/beman/exemplar/CMakeLists.txt
wusatosi, line 1: I'm for this refactoring, but can we make this a single PR, it's unclear exactly what has been changed.

tests/beman/exemplar/CMakeLists.txt
wusatosi: @camio please check this

If this inclusion is sound, can you leave a comment here pointing out why this is needed.
And include this as an explicit exception in the standard. 

Should be here: https://github.com/bemanproject/beman/blob/5c0b8662f9e6ee7b995d5e14f4c690a6aac5ce7a/docs/BEMAN_STANDARD.md?plain=1#L297-L311

@neatudarius might be able to help if we need to update that.

.github/workflows/ci_tests.yml
```diff
@@ -77,36 +77,36 @@ jobs:
...
+            cpp_version: 20
             cmake_args:
               description: "Werror"
               args: "-DCMAKE_CXX_FLAGS='-Werror=all -Werror=extra'"
           - platform:
               description: "Ubuntu GCC"
               os: ubuntu-latest
               toolchain: "cmake/gnu-toolchain.cmake"
-            cpp_version: 17
+            cpp_version: 20
```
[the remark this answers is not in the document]
wusatosi: so we had this discussion last year at sync.
The consensus is to leave minimum support version undefined *but still encourage libraries to support lower version if possible*.
The big argument for 20 was concept support, the big argument for below companies like Adobe is still using 17 and we want their usage experience.
Exemplar can support 17 (hell, it can probably support 11, or even 99!), so to signify this we made exemplar support 17.

I think we can revisit this once 26 comes out.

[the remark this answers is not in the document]
JeffGarland: > The consensus is to leave minimum support version undefined _but still encourage libraries to support lower version if possible_.

Consensus can be changed.  And, technically I'm not *against* leaving the minimum undefined, but *my personal recommendation* to library authors is to not bother with versions below 20.  Concepts, ranges, and a bunch of other things are just too consequential in building a good modern library.  We have limited intellectual capital here and time wasted on supporting older version isn't applied to building a better future.  Remember, we're in the pivot to work on c++29 libraries.  And for 29 don't be surprised if we see people requiring 26 for reflection and contracts.

Let me back this up with some data, however.  According to the 2024 survey, C++20 is not allowed at all in 37% of projects.  The year over year progress from 2023, was in the 6% range -- from 43% not allowed in 2023 to 37%. About 50% answered that they expected to be able to upgrade *this year*.  But conservatively I expect we'd be in the 30% exclusion range and dropping fast. When you note that 8% are still excluded from C++17 this is really only a 22% exclusion.  This is all reasonable as far as I'm concerned -- the slow adopters will eventually get there and our focus here is on the future, not the past.  

As a side note, the barriers to adopting 20+ with an older code base are small in my experience.  We've continued to break very little legacy code so for lots of stuff it's simply changing the compiler flags and moving on.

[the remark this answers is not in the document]
wusatosi: Yeah looks like good points, take this to the sync.
I don't think most libs can support versions below C++20 anyway.

JeffGarland:
> CI fail is not only because proper version is not selected.
> 
> Please check compiler test seris, e.g. llvm 18 CI test. Error message is as following:
> 
> ```
> /home/runner/work/exemplar/exemplar/examples/identity_direct_usage.cpp:6:8: fatal error: module 'beman.exemplar' not found
>     6 | import beman.exemplar;
>       | ~~~~~~~^~~~~
> ```
> 
> See: https://github.com/bemanproject/exemplar/actions/runs/15358684463/job/43222580631?pr=167#step:5:69
> 
> Edit: okay I am mistaken, clang 18 does not support modules.

And in scope PR what I did is check the compiler versions and exclude anything that isn't clang-20 or g++-15 (not yet in CI).  I don't have clang19 setup yet, so I wasn't confident modules would work there.

wusatosi:
> And related note -- scope CI is actually testing other compilers and versions.

Thanks for checking, the bug is introduced in #159  which as far as I know isn't used anywhere yet.

> And in scope PR what I did is check the compiler versions and exclude anything that isn't clang-20 or g++-15 (not yet in CI). I don't have clang19 setup yet, so I wasn't confident modules would work there.

This is out of scope, but if we want usage experience we probably should not be this aggressive on using the latest compilers.

JeffGarland:
> > And in scope PR what I did is check the compiler versions and exclude anything that isn't clang-20 or g++-15 (not yet in CI). I don't have clang19 setup yet, so I wasn't confident modules would work there.

I setup clang++-19 and have now included it in the modules supporting libs since it appears to work fine.

> This is out of scope, but if we want usage experience we probably should not be this aggressive on using the latest compilers.

We can only support modules usage on compilers that are capable of supporting modules (that's probably around 4 at the moment).  The non module users can have a broader range of compilers.  Also, while I understand your advocacy for earlier c++ versions and feature sets, I will continue to remind you that it's not the focus of Beman.  Again, let Boost do that -- they've self selected on that -- we're about the future, not the past.  I expect *much more* in the way of libraries using advanced compilers as reflection and contracts get compiler support.  Which by it's very nature excludes everyone below 26.

wusatosi:
> We can only support modules usage on compilers that are capable of supporting modules (that's probably around 4 at the moment). The non module users can have a broader range of compilers. 

I definitely support this.

> I expect much more in the way of libraries using advanced compilers as reflection and contracts get compiler support. Which by it's very nature excludes everyone below 26.

While I understand supporting 17 maybe not achievable for most beman library (inplace vector doesn't support it either), I do think it is very useful for libraries to support lower C++ versions, e.g. C++20 for right now. In a few years, you can argue for C++23. But if we are always only supporting the latest compilers, latest C++ versions, it will limit us on how we gather usage experience.

> while I understand your advocacy for earlier c++ versions and feature sets, I will continue to remind you that it's not the focus of Beman. 

User feedback is literally in our mission:

> support the efficient design and adoption of the highest quality C++ standard libraries through implementation experience, **user feedback**, and technical expertise.

wusatosi:
I see this is built up on #168 , let's wait for that to get reviewed first. I can go ring up folks with write access to double check it.

nickelpro:
Rebased and removed hack for clang-scan-deps. Only outstanding issues are #168 and a solution for Apple Clang

wusatosi:
> I uh, don't know what to do about that honestly. I'll give it a big think.

Can we have a build option to turn on or off module support? We can default that to off if it's apple clang or too low of a compiler version.

wusatosi:
> Only outstanding issues are #168 and a solution for Apple Clang

I can go ring some folks.

nickelpro:
We can do a dumb thing where we flag-fork the examples to use `import` when it's available and `#include` otherwise, but I don't love that because it's ugly boilerplate. This isn't a problem for the library itself, it only matters for tests and examples where we're more free to use ugly solutions.

dietmarkuehl:
> We can do a dumb thing where we flag-fork the examples to use `import` when it's available and `#include` otherwise, but I don't love that because it's ugly boilerplate. This isn't a problem for the library itself, it only matters for tests and examples where we're more free to use ugly solutions.

You want the exampmes to use `import beman.exemplar;`? We _could_ choose to not compile them on platfforms which don’t support them, possibly with instructions on how to fix things.

While I want to support using modules, I wouldn’t want to force them. I that sense I think it would be good if Beam libraries do support modules but using them via headers should also be possible.

nickelpro:
I want examples and tests to use modules where feasible, yes, because that's the only way to ensure the full interface is being properly exported by the interface unit

wusatosi:
Just to double check, this is meant to be merged into mainline, not only a POC right @nickelpro

nickelpro:
My two cents is that libraries that aim to be a part of the standard have a secondary duty to dogfood the standard. If modules, or concepts, or pattern matching, or whatever, are totally unworkable in the implementation of a standards proposal, that is something the people who go to committee meetings should have first hand experience with.

CMakeLists.txt
```diff
@@ -30,11 +29,78 @@ option(
...
+include(GNUInstallDirs)
 
-add_subdirectory(src/beman/exemplar)
+install(
+    TARGETS beman.exemplar
+    EXPORT beman.exemplar-targets
+    COMPONENT beman.exemplar
+
+    FILE_SET CXX_MODULES
+        DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
```
ClausKlein, line 70: is this intended to install the `cppm` files to `include` tree?

cmake/beman.exemplar-config.cmake
```diff
@@ -0,0 +1,8 @@
+include(${CMAKE_CURRENT_LIST_DIR}/beman.exemplar-targets.cmake)
+
+foreach(comp IN LISTS beman.exemplar_FIND_COMPONENTS)
+    if(beman.exemplar_FIND_REQUIRED_${comp})
+        set(beman.exemplar_FOUND FALSE)
+        return()
+    endif()
```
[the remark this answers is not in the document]
nickelpro: It's hand coded because we don't need any of the features of the generated version. It is not wrong.

We will never use `set_and_check()`.

No Beman project will ever have any `find_package(COMPONENTS)`. Components are only used by find modules, which no Beman project will ever ship. The `check_required_components` loop is used in the scope of such modules.

Arguably we do not need this file at all, we can use the file generated by `install(EXPORT)` directly, but see previous in the linked Beman scope PR at the top of the this PR. The explanation for this loop is listed under the `beman-scope-config.cmake` heading: https://github.com/bemanproject/scope/pull/31

CMakeLists.txt
```diff
@@ -30,11 +29,78 @@ option(
...
+        DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/beman.exemplar
+        NAMESPACE beman::
+        CXX_MODULES_DIRECTORY cxx-modules
+        COMPONENT beman.exemplar
+    )
+endif()
+# gersemi: on
 
 if(BEMAN_EXEMPLAR_BUILD_TESTS)
+    include(CTestUseLaunchers)
```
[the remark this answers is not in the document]
nickelpro: Advanced CTest scripting, see https://github.com/bemanproject/exemplar/pull/109

CMakeLists.txt
nickelpro: Yes.

cmake/beman.exemplar-config.cmake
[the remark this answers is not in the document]
ClausKlein: > Components are only used by find modules, which no Beman project will ever ship.

`Boost` installs `cmake export packages` and supports components?

ClausKlein:
> Only outstanding issues are #168 and a solution for `Apple Clang`

I would forget `apple-clang` which does't fully support POSIX and CXX_MODULES at all.

Also on **OSX** `llvm-19` and `llvm-20` with both I have problems if installed with brew.

see https://gitlab.kitware.com/cmake/cmake/-/issues/25965#note_1575723

I use `gcc-15` installed with brew, which works fine together with `cmake v4.0.2`

CMakeLists.txt
steve-downey: There has been ongoing arguments about where they should go, since they aren't files that one should #include, but there are a two packages that have the module file in include, vulkan and gcc-15 if I recall correctly, so there's some precedent. It doesn't really matter terribly as long as the metadata file for the module points to the location so the module can be compiled in the context of the consumer, with the consumers compiler, flags, etc.

nickelpro: The exist outside the bounds of what is defined by the FHS. In spirit they are closest to header files, being architecture-independent interface descriptions, so they go in `include`.

nickelpro:
> Can we have the module-first be trigger by a compile option?

There's no reason for or advantage to such a division beyond being able to lower the required CMake/compiler versions

nickelpro:
> the actual library, which should be able to be built in 2 modes: modules and non-modules version. We should not assume it's always header only (check - iterator interfaces).

This is just one mode. There is no semantic difference between with and without modules. I'm asking for a use case so I can example this more practically, "I want to be able to do X".

> examples/: here we need to have 2 examples with beman.examplar with modules and the same examples without modules. This is what the users will do.

All the examples will be identical except for a single line, the one that imports the module or includes the header. I do not thing the examples should be replicated to demonstrate a single line can be changed. This can maybe be a configuration option, but honestly what's wrong with just leaving a comment like this PR does?

> tests/: We should have lots of tests with the non-module version of the library to get coverage. Then 1-2 tests which just tests we can build and link with the module version - e.g. call a simple function from the library, just a basic case.

This doesn't do anything, the exact same code is run. The coverage is identical.

ClausKlein:
There is an additional problem with gest, if locally installed, the version is not right checked:

```
bash-5.2$ CXX=g++-15 cmake -S . -B build --fresh  -D CMAKE_CXX_STANDARD=20
-- The CXX compiler identification is GNU 15.1.0
-- Checking whether CXX compiler has -isysroot
-- Checking whether CXX compiler has -isysroot - yes
-- Checking whether CXX compiler supports OSX deployment target flag
-- Checking whether CXX compiler supports OSX deployment target flag - yes
-- Check for working CXX compiler: /usr/local/bin/g++-15 - skipped
-- Found GTest: /usr/local/lib/libgtest.a (Required is at least version "2.17")
-- Looking for __cpp_lib_ranges - found
Examples to be built: identity_direct_usage;identity_as_default_projection
-- Configuring done (2.4s)
-- Generating done (0.0s)
-- Build files have been written to: /Users/clausklein/Workspace/cpp/beman-project/exemplar/build
bash-5.2$ 


FAILED: tests/beman/exemplar/beman.exemplar.tests.identity tests/beman/exemplar/beman.exemplar.tests.identity[1]_tests.cmake /Users/clausklein/Workspace/cpp/beman-project/exemplar/build/tests/beman/exemplar/beman.exemplar.tests.identity[1]_tests.cmake 
: && /usr/local/bin/g++-15  -Wl,-search_paths_first -Wl,-headerpad_max_install_names  tests/beman/exemplar/CMakeFiles/beman.exemplar.tests.identity.dir/identity.test.cpp.o -o tests/beman/exemplar/beman.exemplar.tests.identity  libbeman.exemplar.a  /usr/local/lib/libgtest_main.a  /usr/local/lib/libgtest.a && cd /Users/clausklein/Workspace/cpp/beman-project/exemplar/build/tests/beman/exemplar && /Users/clausklein/.direnv/python-3.13.3/lib/python3.13/site-packages/cmake/data/bin/cmake -D TEST_TARGET=beman.exemplar.tests.identity -D TEST_EXECUTABLE=/Users/clausklein/Workspace/cpp/beman-project/exemplar/build/tests/beman/exemplar/beman.exemplar.tests.identity -D TEST_EXECUTOR= -D TEST_WORKING_DIR=/Users/clausklein/Workspace/cpp/beman-project/exemplar/build/tests/beman/exemplar -D TEST_EXTRA_ARGS= -D TEST_PROPERTIES= -D TEST_PREFIX= -D TEST_SUFFIX= -D TEST_FILTER= -D NO_PRETTY_TYPES=FALSE -D NO_PRETTY_VALUES=FALSE -D TEST_LIST=beman.exemplar.tests.identity_TESTS -D CTEST_FILE=/Users/clausklein/Workspace/cpp/beman-project/exemplar/build/tests/beman/exemplar/beman.exemplar.tests.identity[1]_tests.cmake -D TEST_DISCOVERY_TIMEOUT=5 -D TEST_DISCOVERY_EXTRA_ARGS= -D TEST_XML_OUTPUT_DIR= -P /Users/clausklein/.direnv/python-3.13.3/lib/python3.13/site-packages/cmake/data/share/cmake-4.0/Modules/GoogleTestAddTests.cmake
Undefined symbols for architecture x86_64:
  "testing::internal::MakeAndRegisterTestInfo(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>, char const*, char const*, char const*, testing::internal::CodeLocation, void const*, void (*)(), void (*)(), testing::internal::TestFactoryBase*)", referenced from:
      __static_initialization_and_destruction_0() in identity.test.cpp.o
      __static_initialization_and_destruction_0() in identity.test.cpp.o
      __static_initialization_and_destruction_0() in identity.test.cpp.o
      __static_initialization_and_destruction_0() in identity.test.cpp.o
  "testing::internal::EqFailure(char const*, char const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&, bool)", referenced from:
      testing::AssertionResult testing::internal::CmpHelperEQFailure<int, int>(char const*, char const*, int const&, int const&) in identity.test.cpp.o
      testing::AssertionResult testing::internal::CmpHelperEQFailure<int const*, int const*>(char const*, char const*, int const* const&, int const* const&) in identity.test.cpp.o
  "std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>::find(char, unsigned long) const", referenced from:
      testing::internal::FormatDeathTestOutput(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&) in libgtest.a[2](gtest-all.cc.o)

# # # 

   NOTE: a missing vtable usually means the first non-inline virtual member function has no definition.
  "vtable for std::__1::basic_stringstream<char, std::__1::char_traits<char>, std::__1::allocator<char>>", referenced from:
      std::__1::basic_stringstream<char, std::__1::char_traits<char>, std::__1::allocator<char>>::basic_stringstream[abi:ue170006]() in libgtest.a[2](gtest-all.cc.o)
   NOTE: a missing vtable usually means the first non-inline virtual member function has no definition.
  "vtable for std::__1::basic_ostringstream<char, std::__1::char_traits<char>, std::__1::allocator<char>>", referenced from:
      std::__1::basic_ostringstream<char, std::__1::char_traits<char>, std::__1::allocator<char>>::basic_ostringstream[abi:ue170006]() in libgtest.a[2](gtest-all.cc.o)
   NOTE: a missing vtable usually means the first non-inline virtual member function has no definition.
ld: symbol(s) not found for architecture x86_64
collect2: error: ld returned 1 exit status
bash-5.2$ 


```

ClausKlein:
@nickelpro as you noted in a review, it the project must be build without presets!

```diff
bash-5.2$ git diff CMakeLists.txt
diff --git a/CMakeLists.txt b/CMakeLists.txt
index a241e5e..23563d2 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-cmake_minimum_required(VERSION 3.28)
+cmake_minimum_required(VERSION 3.28...4.0)
 
+set(CMAKE_PROJECT_TOP_LEVEL_INCLUDES cmake/use-fetch-content.cmake)
 project(
     beman.exemplar # CMake Project Name, which is also the name of the top-level
     # targets (e.g., library, executable, etc.).
bash-5.2$ 
```

nickelpro:
I wouldn't generally provide a `<version>` tag to `find_package()` unless you're 100% about the version requirement. The documentation calls this out:

> The `<PackageName>` is the only mandatory argument. The `<version>` is often omitted, and `REQUIRED` should be given if the project cannot be configured successfully without the package.

The correct answer is to rely on the shipped dependency manifest(s) to provision the correct versions of dependencies, trying to verify that after the fact is a redundancy that can only lead to errors.

I didn't write the `FetchContent` machinery and I agree it leaves a lot to be desired. It exists because there was no consensus to give up on `FetchContent`. `FetchContent` _cannot_ work as a general purpose package manager, trying to make it into one is error prone. Better manifest formats are more readable than this ad-hoc one, but honestly it's the best you can do with `FetchContent`.

Downstream packagers shouldn't know anything about project dependency manifests, or the CML, or anything. The Debian maintainer assigned to package your library doesn't want to read any of those things and does not care. They want a ReadMe that gives a set of package names and compatibility ranges, which they will pull from their own Debian repos when packaging your project.

JeffGarland:
With the Bulgaria meeting and everything else going on I'd lost track of this PR -- looks like this is mostly ready to go?

PR #168, opened 2025-05-31
Use testing container

wusatosi:
## Description

Adopts the new testing container in CI with some refactoring.

This should fix the issue described here: https://github.com/bemanproject/exemplar/pull/167#issuecomment-2923947021

## Testing

CI passing.

## Meta

### DIFF
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -113,6 +113,8 @@ jobs:
 
   configuration-test:
     runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/bemanproject/testingcontainers-gcc:14
     strategy:
       fail-fast: false
       matrix:
@@ -126,8 +128,6 @@ jobs:
     name: "CMake: ${{ matrix.args.name }}"
     steps:
       - uses: actions/checkout@v4
-      - name: Setup build environment
-        uses: lukka/get-cmake@latest
       - name: Build and Test
         uses: ./.github/actions/cmake-build-test
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci_tests.yml
```diff
@@ -31,33 +31,32 @@ jobs:
...
+      image: ${{ matrix.presets.compiler && 'ghcr.io/bemanproject/testingcontainers-' }}${{ matrix.presets.compiler }}
     steps:
       - uses: actions/checkout@v4
-      - name: Setup build environment
-        uses: lukka/get-cmake@latest
+      - name: Setup Environment
+        if: ${{ !matrix.presets.compiler }}
+        uses: ./.github/actions/setup-enviornment
         with:
           cmakeVersion: "~3.25.0"
```
JeffGarland, line 44: As noted in the scope reviews, 3.25 is insufficient to support modules 3.28 is required: https://github.com/bemanproject/scope/pull/34/files

nickelpro: Concur, this is just to get the compilers on the right versions

nickelpro (approved):
This does the correct thing and the approach is reasonable.

wusatosi:
Actually I am going to remove the cmake versioning inputs, because this produces inconsistent behavior between testing container and windows/ macos build. In testing container the latest version of cmake and ninja is always pulled.

nickelpro:
It's a non-issue as the CMake behavior is dictated by what the `cmake_minimum_version` is set to, and Ninja doesn't change interfaces almost ever. 

My review is only to say, "yep we're running on the right compiler versions now".

So ya, the CMake/Ninja stuff is whatever works.

camio:
@ednolan, is this PR still relevant given the CI changes you made recently?

ednolan:
No, it's been superseded.

PR #170, opened 2025-06-01
Add .gitattributes file settings the correct line ending settings for…

Jason5480:
… each file according to MS https://code.visualstudio.com/docs/devcontainers/tips-and-tricks#_resolving-git-line-ending-issues-in-containers-resulting-in-many-modified-files

## Description

Add .gitattributes file

## Motivation and Context

When using the devcontainer to develop on a Windows host machine, AND the default settings were used during the git installation (eol=auto). Then there is an issue with line endings that results at files marked as moditied. You can find more here: https://code.visualstudio.com/docs/devcontainers/tips-and-tricks#_resolving-git-line-ending-issues-in-containers-resulting-in-many-modified-files
I followed the tip suggested by MS and this fixed the problem.

## Testing

When develop using devcontainer locally from a host OS that has different line endings than the containers line endings the file sync operation should not change the line endings and as a result the files should not be marked as modified by git.

## Meta

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,9 @@
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+###############################
+# Git Line Endings            #
+###############################
+
+* text=auto eol=lf
+*.{cmd,[cC][mM][dD]} text eol=crlf
+*.{bat,[bB][aA][tT]} text eol=crlf
+*.{vcxproj,vcxproj.filters} text eol=crlf


wusatosi:
I thought GitHub handles this for us?

Can you demonstrate this breaking GitHub based repos?

Jason5480:
> I thought GitHub handles this for us?
> 
> Can you demonstrate line ending breaking GitHub based repos?
> 
> Edit: I actually think I am wrong here. I can't find GitHub side of document on this.

The problem is not on GitHub but on files locally during development. If you installed git on windows and used the default configuration during the installation. 
image
Then, after cloning the exemplar repo, locally you have crlf line endings. If you open the folder in VSCode and use the "Reopen in Container" functionality, then the repo folder is mounted as a volume that is synchronized regularly with the host. If you try to modify a file the linux container uses lf. This causes problems when rsync runs because windows convert lf to crlf the container tries to convert it back to lf... This wastes resources and also makes local git show changed files because of the line endings. In the MS link I posted above there is more info in that.
The issue happening or not, depends on the users' global settings. This could easily be avoided if we set those settings on this repo.

wusatosi:
I see your point, I just don't know how niche this issue is. I am neutral on merging this.

There's an pre-commit based alternative, which I prefer as that does not need another file in the root dir, see: https://github.com/pre-commit/pre-commit-hooks?tab=readme-ov-file#mixed-line-ending

nickelpro:
This is a local configuration problem, don't turn on `autocrlf` and you don't have this problem. There's no reason to ever turn on `autocrlf`, and thus upstreams have no obligation to work around this "problem."

There are dozens, if not hundreds, of potential local configuration issues we can nominally work around upstream, but doing so clutters the repo with irrelevant and sometimes mystifying configuration files and scripts. We should strive to only address universal issues, not issues unique to a particular developer's workflow.

wusatosi:
I still think pre-commit is a good way for us to fend against this.

nickelpro:
Sure but pre-commits are opt-in (I never turn on pre-commits, I'll decide what commands run on my system thanks very much). The downstream dev still needs to do local configuration to turn on pre-commits. So either than can know they need to run `pre-commit install` or they can know they need `git config set core.autocrlf false`.

Either way we're saying "we're not going to solve your local problem automagically, you need to be in control of your local development environment".

wusatosi:
> The downstream dev still needs to do local configuration to turn on pre-commits. So either than can know they need to run `pre-commit install` or they can know they need `git config set core.autocrlf false`.

No it protects us, because we run pre-commit as part of our CI. It will be obvious if you break pre-commit rules, that's where all the review dog comments come from. You have to deal with it one way or another.

> Either way we're saying "we're not going to solve your local problem automagically, you need to be in control of your local development environment".

Yeah the good part of pre-commit is it protects us from this.

nickelpro:
Agreed.

But this PR isn't about protecting us in CI, it's about the dev experience of someone using a Windows dev container.

wusatosi:
> Agreed.
> 
> But this PR isn't about protecting us in CI, it's about the dev experience of someone using a Windows dev container.

You're right, maybe we should request this to be a linter rule then?

Jason5480:
> This is a local configuration problem, don't turn on `autocrlf` and you don't have this problem

If you know what's going on this is an alternative solution of course! But, the autocrlf is the default behavior on Windows and if you don't know what is causing this, it is just a bad User Experience, without any apparent reason to be like that imho. This PR is what MS suggests as a solution and it works in all cases no matter what the local user settings are.

Jason5480:
> This is a local configuration problem, don't turn on `autocrlf` and you don't have this problem

If you know what's going on this is an alternative solution ofcourse ! But, the autocrlf is the default behavior on Windows and if you don't know what's causing this it is a bad User Experience, without an apparent reason to be like that. This PR is what MS suggests so it works in all cases no matter what the 

> @Jason5480 Sounds like devcontainer is part of the problem, can we configure devcontainer to avoid this issue?

As far as I understood, this is caused by rsync that runs from the container and synchronizew the repo folder between the host and the container. However, there is not option for rsync to ignore line endings afaik. The proposed solution by MS is what I did in this MR

wusatosi:
yeah I think I'm honestly netural to both approach, just left to how ichy one feels with a new file in the root directory.
I see the problem and I think it will be nice for us to handle it, as far as I know the pre-commit linter also fixes the file.

JeffGarland:
So this MR seems to be blocked without agreement on moving forward. We probably need a @bemanproject/leads call here, but I'm at least partially pursuaded by the MS link provided at the top of the MR.  While I agree with @nickelpro that we can't solve all local issues this one seems like it's pretty likely to show up. To mitigate the 'cluttering with  irrelevant and sometimes mystifying' config I suggest we have the link to the Microsoft advice here.  I suspect @camio won't be able to comment for awhile...

camio:
@Jason5480, could you add a reference to the Microsoft article as a comment in the `.gitattributes` file? If so, I'm good with approving this.

camio (approved):

.gitattributes
```diff
@@ -0,0 +1,15 @@
...
+# It ensures consistent line endings across different operating systems.
+# For more information, see: https://git-scm.com/docs/gitattributes
+# Set the default behavior to automatically normalize line endings and convert them to LF in the repository.
+# This fixes possible consistency issues for cross-platform projects when developing in a devcontainer.
+# For more info please check https://code.visualstudio.com/docs/devcontainers/tips-and-tricks#_resolving-git-line-ending-issues-in-containers-resulting-in-many-modified-files.
+
+* text=auto eol=lf
+*.{cmd,[cC][mM][dD]} text eol=crlf
+*.{bat,[bB][aA][tT]} text eol=crlf
+*.{vcxproj,vcxproj.filters} text eol=crlf
```
neatudarius, line 15: You should also add this inside cookicutter - CC: @ednolan

neatudarius (approved):

ednolan (changes requested):
I just discovered that this option causes no end of grief when checking out repositories that contain shell scripts. [This](https://stackoverflow.com/a/39527986) Stack Overflow answer contains an example:

> With Windows or DOS-style line endings , each line is terminated with a Carriage Return followed by a Line Feed character. You can see this otherwise invisible character in the output of `cat -v yourfile`:

```
$ cat -v yourfile
#!/bin/bash^M
^M
cd "src"^M
npm install^M
^M
cd ..^M
./tools/nwjs-sdk-v0.17.3-osx-x64/nwjs.app/Contents/MacOS/nwjs "src" &^M
```

> In this case, the carriage return (`^M` in caret notation or `\r` in C escape notation) is not treated as whitespace. Bash interprets the first line after the shebang (consisting of a single carriage return character) as the name of a command/program to run.

>  - Since there is no command named `^M`, it prints : command not found
>  - Since there is no directory named `src^M`, it prints : No such file or directory
>  - It passes `install^M` instead of install as an argument to npm which causes npm to complain.

This seems like an excellent reason not to allow this change. Beman library developers will want to be able to provide shell scripts, and users and maintainers will want to be able to run them under WSL in Windows.

Jason5480:
> Actually, I just realized that what I wrote above shouldn't really be a problem in practice because you can exclude shell scripts using `.gitattributes`. But the fact that it's something we even have to think about makes me suspicious of this feature.

In the proposed .gitattributes the .sh files will be committed with lf linendings as per line `* text=auto eol=lf`. If the file was present in the repository at the time you committed your *.sh file. Otherwise, your global settings are used. If you had different settings when you committed the file you have to renormalize the line endings like that `git add --renormalize`.  I have the impression that what you experience would have been prevented if this file was present at the time to the *.sh commit.

PR #179, opened 2025-06-10
Add cookiecutter template

ednolan:
This commit introduces a 'cookiecutter' directory that users can use to spin off new projects using exemplar as a template. After invoking cookiecutter on the repository, users are prompted for a new project name, directory name, C++ version, paper, owner, and description, and these values are filled into the template.

It also adds two new checks into CI: that, when the template is filled in with the values for exemplar itself, the result is identical to exemplar; and, that if a name other than "exemplar" is specified for the project name, then the string "exemplar" does not appear in the output when the template is filled in.

These CI checks can be tested by maintainers of exemplar without
needing to update the pull request by running the script
check_cookiecutter.sh.


cookiecutter/{{cookiecutter.directory_name}}/.github/CODEOWNERS
```diff
@@ -0,0 +1,23 @@
...
+# .github/workflows/ @wusatosi # Add other project owners here
+#
+# Devcontainer:
+# .devcontainer/ @wusatosi # Add other project owners here
+#
+# Pre-commit:
+# .pre-commit-config.yaml @wusatosi # Add other project owners here
+# .markdownlint.yaml @wusatosi # Add other project owners here
+
+*  @bretbrownjr @camio @dietmarkuehl @neatudarius @steve-downey @wusatosi @bemanproject/core-reviewers
```
JeffGarland, line 23: I think there's overlap here? Several of these people are in core-reviewers?

ednolan: This file is identical to the existing `examplar/.github/CODEOWNERS` file, so if that's an issue then it wasn't introduced by this commit

neatudarius: Please continue with current line, the group is a fallback https://github.com/bemanproject/beman/blob/main/docs/BEMAN_STANDARD.md#repositorycodeowners, and it will not pop people already tagged.

JeffGarland: Fair enough - redundancy is trigging me, but fine :)

neatudarius (approved):

ClausKlein:
@ednolan I miss some information what is to do if a developer change something on this repo?

i.e. changes on `CMakeLists.txt`

ednolan:
You need to make all your changes twice-- e.g., once in `/CMakeLists.txt` and once in `/cookiecutter/{{cookiecutter.project_name}}/CMakeLists.txt`-- and make sure any use of the word `exemplar`/`EXEMPLAR` is templated, i.e. replaced with `{{cookiecutter.project_name}}` or `{{cookiecutter.project_name}}.upper()`, respectively.

ClausKlein:
> You need to make all your changes twice

and what is the real benefit of tis cookie cutter?

ednolan:
All of those examples are either in the CI script that checks the cookiecutter template, or in the `infra` beman-submodule, which is excluded from templating and contains the string `exemplar` for unrelated reasons. There is a CI check that ensures that the string `exemplar` does not appear elsewhere in the cookiecutter directory.

ClausKlein:
But than this template is only usable for `beman` project?

ednolan:
Cookiecutter will also prompt you for the `owner` field, and you can input something other than `bemanproject`. However, that's not currently tested in CI because it's more difficult to test for that than testing for `exemplar`, since, unlike with `exemplar`, there are various instances of the string `bemanproject` that shouldn't be replaced by the template. So it's currently best-effort.

ClausKlein:
sorry to say, but that is  the last contibution to this project, to mutch work!

Issue #192, opened 2025-06-30
GitHub detects wrong language for repos


wermos:
## Describe the deficiency

GitHub currently shows Python as the main programming language used in the repo. The affected repos are the ones which are smaller features, hence requiring fewer lines of C++. Here's a list:
- [`exemplar`](https://github.com/bemanproject/exemplar/)
- [`indices_view`](https://github.com/bemanproject/indices_view)
- [`copyable_function`](https://github.com/bemanproject/copyable_function)
- [`cstring_view`](https://github.com/bemanproject/cstring_view) - This one is low priority as it is still heavily WIP.

Some other repos in the beman organization are suffering from the same issue, but with CMake. The cause is similar; it's because of the large amount of CMake boilerplate in those repos.

- [`monadics`](https://github.com/bemanproject/monadics)
- [`dump`](https://github.com/bemanproject/dump)

Both of these are lower priority fixes because `monadics` has not received any commits for a couple of months, and `dump` is marked as archived.

## Expected Behavior

GitHub should show C++ as the main language of the exemplar repo and all repos derived from this template.

## Additional Discussions

To fix these issues, we need to add a `.gitattributes` file to help [GitHub Linguist](https://github.com/github-linguist/linguist/tree/main) (GitHub's repo language detection tool) correctly identify the language of the repo.

wermos:
Closing this as it was addressed by #193.

PR #193, opened 2025-06-30
Add .gitattributes to help GitHub Linguist

wermos:
## Description

This PR addresses the issues described in #192.

I merged this to `main` on my fork and the results are as follows:
image

## Meta

I'm not so sure about whether this should be merged to exemplar, considering the fact that `cookiecutter/` is a bona fide part of this repo. But I think that this `.gitattributes` file should definitely be there in repos generated from the template, like `indices_view`, `copyable_function`, etc.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,5 @@
 infra/** linguist-vendored
 cookiecutter/** linguist-vendored
+*.bib -linguist-detectable
+*.tex -linguist-detectable
+papers/* linguist-documentation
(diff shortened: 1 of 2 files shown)


.gitattributes
```diff
@@ -0,0 +1,2 @@
+infra/** linguist-vendored
```
neatudarius, line 1: Can you also add these https://github.com/bemanproject/optional/blob/main/.gitattributes?

We should also exclude LaTeX

ednolan (approved):

.gitattributes
wermos: Done.

> We should also exclude LaTeX

How should we go about that? Should I exclude all `*.tex` files?

neatudarius: Yes, now it's good

neatudarius (approved):

PR #199, opened 2025-07-02
Run preset tests for GCC and Clang on containers2 images

ednolan:


ednolan:
This was a test of CI not intended to be merged.

Issue #182, opened 2025-06-17
cmake -LH needs source dir specified


wermos:
In the README, there is [a section](https://github.com/bemanproject/exemplar/blob/aa60f6d8afd3abaf1b89276c680e98b73ab6c700/README.md#project-specific-configure-arguments) that takes about running `cmake -LH | grep "BEMAN_INDICES_VIEW" -C 2`.

I tried running that on my local machine and I got this warning:
```
CMake Warning:
  No source or binary directory provided.  Both will be assumed to be the
  same as the current working directory, but note that this warning will
  become a fatal error in future CMake releases.
```

Running `cmake -LH -S . | grep "BEMAN_INDICES_VIEW" -C 2` (note the added `-S .` flag to the CMake command fixes the issue.

My CMake version is 4.0.2, and I am on Ubuntu 24.10.

camio:
Thanks for creating this issue. Would you be willing to create a PR with the fix?

wermos:
Sure thing.

Issue #139, opened 2025-03-10
Adopt suggestion from post-merge review for new README


wusatosi:
Evaluate and adopt suggestions to README file by david post merge.

https://github.com/bemanproject/exemplar/pull/111#pullrequestreview-2659237096

wermos:
Most of the post-review suggestions were addressed by #232. To see a detailed list of what is left to do, see [this comment](https://github.com/bemanproject/exemplar/pull/232#issuecomment-3092425829) in that PR.

wusatosi:
Thanks @wermos ! Your PR should be sufficient to close this issue!!! Thank you so much for contributing

PR #232, opened 2025-07-19
Update README

wermos:
Addresses #182 by updating the relevant CMake command.

I also took the liberty of addressing #139. Even though the issue is already assigned, there has been no work on it since its creation, so I decided to kill two birds with one PR.

### DIFF
--- a/cookiecutter/{{cookiecutter.project_name}}/README.md
+++ b/cookiecutter/{{cookiecutter.project_name}}/README.md
@@ -86,23 +86,22 @@ Full runnable examples can be found in [`examples/`](examples/).
 
 This project requires at least the following to build:
 
-* C++17
-* CMake 3.25
+* A C++ compiler that conforms to the C++17 standard or greater
+* CMake 3.25 or later
 * (Test Only) GoogleTest
 
-You can disable building tests by setting cmake option
+You can disable building tests by setting CMake option
 [`BEMAN_{{cookiecutter.project_name.upper()}}_BUILD_TESTS`](#beman_{{cookiecutter.project_name}}_build_tests) to `OFF`
 when configuring the project.
 
-Even when tests are being built and run, some will not be compiled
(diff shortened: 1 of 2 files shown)


steve-downey (approved):

ednolan (approved):

neatudarius (approved):

cookiecutter/{{cookiecutter.project_name}}/README.md
```diff
@@ -1,17 +1,17 @@
...
+![Library Status](https://raw.githubusercontent.com/bemanproject/beman/refs/heads/main/images/badges/beman_badge-beman_library_under_development.svg) ![Continuous Integration Tests](https://github.com/bemanproject/exemplar/actions/workflows/ci_tests.yml/badge.svg) ![Lint Check (pre-commit)](https://github.com/bemanproject/exemplar/actions/workflows/pre-commit.yml/badge.svg) [![Coverage](https://coveralls.io/repos/github/bemanproject/exemplar/badge.svg?branch=main)](https://coveralls.io/github/bemanproject/exemplar?branch=main) ![Standard Target](https://github.com/bemanproject/beman/blob/main/images/badges/cpp29.svg)
```
JeffGarland, line 8: sorry -- I think this is not correct, it's regression for cookiecutter - @ednolan do you agree?  There's more below as well

ednolan: I agree-- this was caught by CI. This shouldn't land until this is fixed and CI is passing.

wermos: Thanks for the diff, this makes it easy to see what the required change is. I remembered to update cookiecutter, but I didn't realize that a wholesale copy-paste wouldn't work.

wermos: I've fixed the cookiecutter README.

ednolan (approved):

PR #235, opened 2025-07-19
Apply RELEASE.GODBOLT_TRUNK_VERSION

neatudarius:


JeffGarland:
698 files changed! oh no....

JeffGarland:
so is it possible to say 'SKIPPED' instead of ERROR/ WARNING on the output from tidy?  As a user it's confusing that the tool is telling me there's an ERROR and then the check passes...

neatudarius:
> 698 files changed! oh no....

It was a mistake, I added a build venv directory.. removed.

neatudarius:
@ednolan , what should I put as a placeholder for that link? https://github.com/bemanproject/exemplar/actions/runs/16394014987/job/46323682814?pr=235

cookiecutter/{{cookiecutter.project_name}}/README.md
```diff
@@ -5,7 +5,7 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
...
+![Library Status](https://raw.githubusercontent.com/bemanproject/beman/refs/heads/main/images/badges/beman_badge-beman_library_under_development.svg) ![Continuous Integration Tests](https://github.com/{{cookiecutter.owner}}/{{cookiecutter.project_name}}/actions/workflows/ci_tests.yml/badge.svg) ![Lint Check (pre-commit)](https://github.com/{{cookiecutter.owner}}/{{cookiecutter.project_name}}/actions/workflows/pre-commit.yml/badge.svg) [![Coverage](https://coveralls.io/repos/github/{{cookiecutter.owner}}/{{cookiecutter.project_name}}/badge.svg?branch=main)](https://coveralls.io/github/{{cookiecutter.owner}}/{{cookiecutter.project_name}}?branch=main) ![Standard Target](https://github.com/bemanproject/beman/blob/main/images/badges/cpp29.svg) [![Compiler Explorer Example](https://img.shields.io/badge/Try%20it%20on%20Compiler%20Explorer-grey?logo=compilerexplorer&logoColor=67c52a)](https://godbolt.org/z/{{cookiecutter.project_name}})
```
JeffGarland, line 8: idk that this actually works -- we can replace the name, but it's going to fail until a human generates a linke for this. Also at the start these links won't be valid for all libraries until we request them in godbolt.  This almost seems like we shouldn't try to enforce this.

neatudarius:
> so is it possible to say 'SKIPPED' instead of ERROR/ WARNING on the output from tidy? As a user it's confusing that the tool is telling me there's an ERROR and then the check passes...

I'll do that in https://github.com/bemanproject/infra/issues/147. Thanks for suggestion!

cookiecutter/{{cookiecutter.project_name}}/README.md
neatudarius: That means for similar situations, we'll have recommandation that is implemented in the tool, but exemplar fails to conform.

Ideally, we should add the badge just to the exemplar, and not to the cookiecutter tree. @ednolan , is that possible?

ednolan:
> I would like to have a broken link into generated repo, because beman-tidy will detect if the suffix is not [a-zA-Z0-9]+ matched. I also tried with TODO_REPLACE_ME.

Our current CI system for cookiecutter won't support this, because it will fail if the output of cookiecutter does not exactly match exemplar.

cookiecutter/{{cookiecutter.project_name}}/README.md
ednolan: No, CI's not smart enough to add those exclusions currently.

ednolan: I thought of a solution. We can add a "godbolt_example_link" parameter to cookiecutter that the user gets prompted to fill in. Our CI will explicitly set it to `https://godbolt.org/z/4qEPK87va` so that the check that it matches `exemplar` passes, but the default can just be something like `www.example.com`, and users will get an interactive prompt when they're stamping out the repo so they'll be aware that they should fill it in properly.

ednolan: Pushed a commit to this branch that implements this.

neatudarius: I'm OK with proposed solution.

PR #241, opened 2025-08-11
Call beman_install_library

bretbrownjr:
Problem
-------

The complexity required per project to configure install logic is too high. It is also too decentralized, making it difficult to improve the various Beman repos packaging logic over time.

Solution
--------

Call `beman_install_library` instead. This function will provide reusable installation logic based on Beman Standards.

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,7 +242,12 @@ To configure, build and test the project with extra arguments,
 you can run this set of commands.
 
 ```bash
-cmake -B build -S . -DCMAKE_CXX_STANDARD=20 # Your extra arguments here.
+cmake \
+  -B build \
+  -S . \
+  -DCMAKE_CXX_STANDARD=20 \
+  -DCMAKE_PREFIX_PATH=./infra/cmake \
+  # Your extra arguments here.
 cmake --build build
 ctest --test-dir build
 ```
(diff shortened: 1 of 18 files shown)


bretbrownjr:
This requires a sync of the infra-vendored files, so this PR cannot merge until https://github.com/bemanproject/infra/pull/171 is merged.

bretbrownjr:
@neatudarius @wusatosi @ednolan What's the process for updating the cookiecutter subdir?

Or is everyone just making every change twice, once in jinja and again in the actual exemplar code?

ednolan:
> is everyone just making every change twice, once in jinja and again in the actual exemplar code?
That is the current process, unfortunately.

bretbrownjr:
Note that the cookiecutter consistency check should probably diff the `infra/.beman-submodule` file, but *not* anything else in the `infra` subdir. Note that [`cookiecutter` has post-generation hooks](https://cookiecutter.readthedocs.io/en/1.7.2/advanced/hooks.html). Using those and keeping the cookiecutter subdir leaner probably makes more sense than having two copies of `infra` in the exemplar repo.

bretbrownjr:
I propose we consider the cookiecutter breakage off topic for this PR. It looks like we need more work on the cookiecutter consistency checks, particularly when a `beman-submodule update` bump is in the picture. As I noted above, I expect the right approach is to make the cookiecutter subdir remove the `infra/` contents.

ednolan:
Thanks for the tip about the post-generation hooks, I wasn't aware of that feature. I agree that using that to dedup `infra` here sounds like a good idea.

ednolan (approved):

bretbrownjr:
The matching `infra` change is merged. The commit SHA in the `.beman-submodule` files uses the latest commit from `infra`.

PR #244, opened 2025-09-02
FIX: traditional CMake invocation failed

domibel:
Add -DCMAKE_PROJECT_TOP_LEVEL_INCLUDES=./infra/cmake/use-fetch-content.cmake to be able to download gtest during the configuration stage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,6 +247,7 @@ cmake \
   -S . \
   -DCMAKE_CXX_STANDARD=20 \
   -DCMAKE_PREFIX_PATH=./infra/cmake \
+  -DCMAKE_PROJECT_TOP_LEVEL_INCLUDES=./infra/cmake/use-fetch-content.cmake \
   # Your extra arguments here.
 cmake --build build
 ctest --test-dir build


nickelpro:
`use-fetch-content.cmake` is a mechanism of last resort for the presets. It exists for users who want an "out-of-the-box" build without concern for how building C++ or downloading dependencies work. This is an uncommon workflow to support in open source C++, which is why we have an internal Beman-ism abstraction instead of a general purpose industry-standard tool.

Traditional CMake requires the user to provision the environment as described in the build's documentation. This is the expected mechanism. While other projects may choose to transparently bootstrap dependencies using standard package managers under some conditions, none provide something like `use-fetch-content.cmake`.

In short, those uninterested in the mechanics of the build should use the presets, those who want a "traditional" CMake build should provision the environment the same way they do any other CMake project, and those clever enough to find `use-fetch-content.cmake` don't need the documentation.

I'm opposed to publicly documenting this detail in the ReadMe. Open to discussion on it.

ednolan:
`use-fetch-content.cmake` is already documented. If your environment is missing `GTest` all you need to do is scroll down to the next section:
Screenshot 2025-09-22 at 3 55 24 PM

PR #252, opened 2025-09-14
Fix cron for pre-commit auto-update

ednolan:
Previously, the cron schedule intended for use by pre-commit auto-update did not match the if condition for that job. This fix brings them into alignment.


cookiecutter/{{cookiecutter.project_name}}/.github/workflows/ci_tests.yml
wusatosi, line 1: I realized this is a bit too late for reviews, but maybe a better thing to do here is to generate a cron descriptor based on the name of the repo so we don't have every repo having their scheduled CI run at the same time?

PR #254, opened 2025-09-14
Reorganize pre-commit autoupdate into separate workflow file

ednolan:
Previously, the scheduling logic in ci_tests.yml was broken due to trying to juggle multiple crons intended for different purposes. This commit addresses the issue by moving the pre-commit autoupdate job into a separate workflow file with its own cron.


ednolan:
Superseded by https://github.com/bemanproject/exemplar/pull/260

PR #262, opened 2025-09-29
WIP Draft: Convert exemplar into an INTERFACE library

ednolan:
This commit eliminates the src/ directory, moving
beman.examplar-config.cmake.in into the top level directory, moving the contents of src/beman/exemplar/CMakeLists.txt into the top level CMakeLists.txt, removing identity.cpp, and changing beman.exemplar from STATIC/SHARED into INTERFACE.

This commit is currently just intended to be a jumping-off point for discussion. What I would like to do if we are okay with this direction is to preserve the option for users to create STATIC/SHARED libraries by configuring cookiecutter differently. That feature is not present yet here.


ednolan:
I opened a thread on Discourse here: https://discourse.bemanproject.org/t/converting-exemplar-to-a-cmake-interface-type-library/527

CC @bretbrownjr @nickelpro

CMakeLists.txt
```diff
@@ -24,9 +24,24 @@ option(
...
+add_library(beman.exemplar INTERFACE)
+add_library(beman::exemplar ALIAS beman.exemplar)
+
+target_sources(
+    beman.exemplar
+    PUBLIC
+        FILE_SET HEADERS
+            BASE_DIRS "${CMAKE_CURRENT_SOURCE_DIR}/include"
+            FILES
+                "${CMAKE_CURRENT_SOURCE_DIR}/include/beman/exemplar/identity.hpp"
```
camio, line 36: It's interesting how the header list moves to the top-level directory when it's a header-only library. This has two drawbacks:

1. A header-only -> with-source-file transition is more dramatic.
2. For a repository with multiple libraries, the top-level CMake file would grow quite large and we lose the "one CMakeLists.txt file per library" separation of concerns.

What are your thoughts on peeing the `src/beman/exemplar` directory and having it contain only a `CMakeLists.txt` that builds the target?

JeffGarland:
It would be good to move this forward asap.  I think you're going to need to rectify with main -- in particular this PR which got merged:

https://github.com/bemanproject/exemplar/pull/275

CMakeLists.txt
```diff
@@ -27,7 +27,19 @@ option(
...
+    PUBLIC FILE_SET HEADERS BASE_DIRS "${CMAKE_CURRENT_SOURCE_DIR}/include"
+)
+
+set_target_properties(beman.exemplar PROPERTIES VERIFY_INTERFACE_HEADER_SETS ON)
+
+add_subdirectory(include/beman/exemplar)
+
+include(CTest)
 
 find_package(beman-install-library REQUIRED)
```
JeffGarland: See my comment about merging from main -- this should be include not find_package

wusatosi:
Let me try to trigger the new GitHub actions script on my local fork, will approve if that passes.

Issue #274, opened 2025-12-25
Exemplar various updates


JeffGarland:
Some of the basics in examplar are out of date -- like the cmake version to support.

PR #275, opened 2025-12-25
#274 - small tweaks in cmake

JeffGarland:
When reviewing the take_before repo I found some issues.

### DIFF
--- a/cookiecutter/{{cookiecutter.project_name}}/CMakeLists.txt
+++ b/cookiecutter/{{cookiecutter.project_name}}/CMakeLists.txt
@@ -25,7 +25,7 @@ option(
 )
 
 # for find of beman-install-library
-set(CMAKE_PREFIX_PATH ${CMAKE_CURRENT_SOURCE_DIR}/infra/cmake)
+include(infra/cmake/beman-install-library-config.cmake)
 
 add_subdirectory(src/beman/{{cookiecutter.project_name}})
 


cookiecutter/{{cookiecutter.project_name}}/CMakeLists.txt
```diff
@@ -1,13 +1,13 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-cmake_minimum_required(VERSION 3.25)
+cmake_minimum_required(VERSION 3.28...4.2)
```
JeffGarland: 3.28 is minimum for modules -- we should have that as minimum now

neatudarius (approved):
LGTM

cookiecutter/{{cookiecutter.project_name}}/CMakeLists.txt
JeffGarland: Well I think that's one persons view, not necessarily the consensus.  And for me it should be a goal for a header only library to support as it's a frequent way to consume header only libraries. cmake is 100% our standard for bulding, packaging, and ci -- but that shouldn't *prevent* other common use cases.  Let's say I had a Bazel project and I"m going to do the work to integrate it.  For header only that means simply copy the files and putting them somewhere and doing and include.  We might even want it for our own purposes.  If I wanted to make an 'all the views super project' for beman I'd just need to clone the headers...

PR #281, opened 2026-01-04
Fix required CMake version in README

ednolan:
We require 3.28 but the README stated 3.25 instead.


cookiecutter/{{cookiecutter.project_name}}/README.md
```diff
@@ -91,7 +91,7 @@ Full runnable examples can be found in [`examples/`](examples/).
 This project requires at least the following to build:
 
 * A C++ compiler that conforms to the C++17 standard or greater
-* CMake 3.25 or later
+* CMake 3.28 or later
```
JeffGarland: Thanks for this -- I missed this in my PR!

PR #389, opened 2026-05-04
Change stamp.sh into stamp.py

### LINKED ISSUE #388: Windows users cannot run stamp.sh without WSL
I think exemplar should
1. Have a unified `stamp.py` for all platforms, since the script uses python anyway

OR

2. Have both `stamp.sh` and `stamp.ps1` (for windows powershell)

AndreiDurlea:
Fixes #388 (implements option 1) and allows windows users to run the stamp script easily. Also allows doing certain CI jobs that test on cookiecutter's output (bemanproject/beman-tidy#312) on windows-latest


AndreiDurlea:
@ednolan @neatudarius what do we think? Do i get the green light to merge this?

btw kept git history clean
```
commit f1cad195e646e4ae4b19327305843596b19aeeba (HEAD -> 388-windows-users-cannot-run-stampsh-without-wsl, origin/388-windows-users-cannot-run-stampsh-without-wsl)
Author: Andrei Durlea <andrei.durlea.telefon@gmail.com>
Date:   Tue May 5 01:50:57 2026 +0300

    ci: Add @neatudarius to root CODEOWNERS

    Forgotten change from #387

commit b2cad0f866a746176ec4b9285cd71f6168e204f1
Author: Andrei Durlea <andrei.durlea.telefon@gmail.com>
Date:   Tue May 5 01:45:28 2026 +0300

    Change stamp.sh into stamp.py
```

AndreiDurlea:
Updated this with main just now, everything is passing. Could be useful for users on other platforms as well. Is this a wanted change? @neatudarius @ednolan

wusatosi (commented):
The bash script -> python transcription is sound.
The motivation is right.

@ednolan , @camio , @neatudarius care to chime in?

ClausKlein:
In general, we should reduce the use of bash scripts. So it is the right way to the future.

But note this file may be obsolete if we switch from `cookiecutter` to `copier`!

camio (approved):

ednolan (changes requested):
I want to hold this commit back until we can reconcile it with the copier changes. Having both of them in flux at the same time is going to result in the need for rebasing one on the other which I want to wrap my head around before we merge it.

AndreiDurlea:
> I want to hold this commit back until we can reconcile it with the copier changes. Having both of them in flux at the same time is going to result in the need for rebasing one on the other which I want to wrap my head around before we merge it.

Alright, I'll wait for the copier stuff to hit main before coming back to this PR. 

PS: when i first made this PR, i had an asus, so I cared more about this kinda issue =). Now that I own a macbook (finally!), such issues might no longer catch my eye - we need to get windows users on the team=)

wusatosi:
> PS: when i first made this PR, i had an asus, so I cared more about this kinda issue =). Now that I own a macbook (finally!), such issues might no longer catch my eye - we need to get windows users on the team=)

Congrats on your new laptop! We can test this using the windows runners.

PR #409, opened 2026-05-27
Use absolute path to toolchain files

ClausKlein:
see too https://discourse.cmake.org/t/how-to-force-that-cmake-toolchain-file-is-an-absolute-file-path/15670/3?u=clausklein

### DIFF
--- a/cookiecutter/{{cookiecutter.project_name}}/CMakePresets.json
+++ b/cookiecutter/{{cookiecutter.project_name}}/CMakePresets.json
@@ -8,8 +8,10 @@
       "binaryDir": "${sourceDir}/build/${presetName}",
       "cacheVariables": {
         "CMAKE_CXX_STANDARD": "{{cookiecutter.minimum_cpp_build_version}}",
-        "CMAKE_EXPORT_COMPILE_COMMANDS": "ON",
-        "CMAKE_PROJECT_TOP_LEVEL_INCLUDES": "./infra/cmake/use-fetch-content.cmake"
+        "CMAKE_CXX_EXTENSIONS": false,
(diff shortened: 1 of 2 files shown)


cookiecutter/{{cookiecutter.project_name}}/CMakePresets.json
```diff
@@ -8,8 +8,10 @@
       "binaryDir": "${sourceDir}/build/${presetName}",
       "cacheVariables": {
         "CMAKE_CXX_STANDARD": "{{cookiecutter.minimum_cpp_build_version}}",
-        "CMAKE_EXPORT_COMPILE_COMMANDS": "ON",
-        "CMAKE_PROJECT_TOP_LEVEL_INCLUDES": "./infra/cmake/use-fetch-content.cmake"
+        "CMAKE_CXX_EXTENSIONS": true,
```
ednolan: > `"CMAKE_CXX_EXTENSIONS": true,`

I think we actually want to turn off extensions in the Beman project. We don't want standards proposals accidentally relying on nonstandard features.

ednolan (changes requested):

cookiecutter/{{cookiecutter.project_name}}/CMakePresets.json
ClausKlein: Ok, I am happy, because at least this works too with this preset on **OSX**:
```bash
cmake --preset gcc-release --fresh --log-level=VERBOSE -D BEMAN_EXEMPLAR_USE_MODULES=1
```

ednolan (approved):

PR #414, opened 2026-07-15
Support hyphenated project names in the cookiecutter template

steve-downey:
## Summary

Stamping a project whose name contains a hyphen (e.g. `my-lib`) previously produced invalid C++ and CMake: `namespace beman::my-lib`, `BEMAN_MY-LIB_USE_MODULES`, and so on. Additionally, three generated files contained a hardcoded `BEMAN_EXEMPLAR_USE_MODULES` regardless of the project name.

- Fix hardcoded `BEMAN_EXEMPLAR_USE_MODULES` in the generated `identity.hpp`, `identity.test.cpp`, and tests `CMakeLists.txt`. It slipped through because `check_cookiecutter.sh` greps for untemplated `exemplar` case-sensitively; the grep is now case-insensitive.
- Add a derived `__project_ident` cookiecutter variable (`project_name` with `-` rewritten to `_`), used wherever the name must be a legal identifier or a source path: the `beman::<ident>` namespace, `beman.<ident>` module/target/package names, `BEMAN_<IDENT>_*` macros and options, and the `include/beman/<ident>/` and `tests/beman/<ident>/` layout (template directories and the `<ident>.hpp`/`<ident>.cppm` files are renamed accordingly, and the post-gen hook follows).
- `project_name` keeps its hyphen where it names the repo or port: top-level directory, GitHub URLs and badges, `REPO bemanproject/<name>`, vcpkg `beman-*` port names, and README prose.
- `check_cookiecutter.sh` now also stamps with a hyphenated sentinel name (`RLZrmX9-NfS`) and fails if the hyphenated form leaks into `beman.`/`beman::`/`beman/` contexts, uppercase identifiers, or any file or directory name.

## Testing

- `check_consistency` passes: stamping with `project_name="exemplar"` still reproduces this repository byte-for-byte, so the change is a no-op for non-hyphenated names.
- Extended `check_templating` passes.
- Stamped a project as `my-lib` and built it with GCC 15: header-only configuration builds and links, and `-DBEMAN_MY_LIB_USE_MODULES=ON` compiles `export module beman.my_lib;` and links `libbeman.my_lib.a`.


ednolan:
> Fix hardcoded BEMAN_EXEMPLAR_USE_MODULES in the generated identity.hpp, identity.test.cpp, and tests CMakeLists.txt. It slipped through because check_cookiecutter.sh greps for untemplated exemplar case-sensitively; the grep is now case-insensitive.

This part is logically separate and should be a separate pull request.

For the hyphenated project names, what is the motivation? Project names in beman repositories are supposed to use underscores, unless it's an infra repository like infra-containers. If it's a sandbox repository like sandbox-slice_view, the cookiecutter stamp should still use "slice_view" as the project name; the "sandbox-" prefix should apply to the repository URL in GitLab but putting it in the CMake, header names, etc., is just clutter.

steve-downey:
Came across it while trying to separate out project name from the directory name the project is in. 
And I work with too many different naming rules and can never remember which ones are required and which ones are forbidden for particular ecosystems and tools. 
I'll factor out the stray EXEMPLAR into a different commit.

PR #416, opened 2026-07-15
fix: disable cmake.default and passive_targets in cookiecutter tidy config

neatudarius:
## Summary

Unblocks `Run on beman.exemplar` for beman-tidy PRs implementing `cmake.default` and `cmake.passive_targets`.

Temporarily disables both checks in the cookiecutter `.beman-tidy.yaml` for stamped libraries. Each disabled rule has a TODO explaining why. No CMake changes — module support stays intact.

## Test plan

- [x] Cookiecutter stamp + `beman-tidy --require-all` passes
- [x] beman-tidy `run-on-exemplar` CI green after merge
- [x] Exemplar module CI still passes

### DIFF
--- a/.beman-tidy.yaml
+++ b/.beman-tidy.yaml
@@ -5,6 +5,7 @@
 # https://github.com/bemanproject/beman-tidy/blob/main/README.md
 
 disabled_rules:
+  # TODO: exemplar is the template repository, not a real Beman library — its README title is not meant to match a library name.
   - readme.title
 ignored_paths:
   - infra/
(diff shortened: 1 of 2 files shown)


ednolan (changes requested):
This breaks module support (as picked up by CI). We need to think about this more.

neatudarius:
> This breaks module support (as picked up by CI). We need to think about this more.

Agree. It's just an experimental change. It's OK that you blocked it, but I wasn't going to put this to review (anyways, not merge without discussion).

I propose just to keep the checks disabled inside cookiecutter. And we'll discuss next week more. 

@ednolan , it is OK this way?

ednolan:
Sounds good

neatudarius:
I'm doing admin merge since Eddie also gave the like.

PR #418, opened 2026-07-26
Default C++ standard version to C++20 (#397)

### LINKED ISSUE #397: exemplar should consistently use c++20 a default cxx version
Per the conversation here https://discourse.bemanproject.org/t/potential-policy-change-making-c-20-the-minimum-library-implementation/581/5 we'd like to encourage c++20 going forward as the base beman version. Assuming the policy decision is to make 17 an option for the library author we still want to default to 20. I notice there's at least one case in cookie cutter stamp.sh that defaults to 17.

wusatosi:
## Summary 
Change the default C++ standard version to C++20 per #397  .

CC @JeffGarland 

## Note to reviewers
- Updated cookie cutter
- Updated docs (CONTRIBUTING.md)
- Updated CMakePresets

## Need help
I have no idea why vcpkg ci is failing.


cookiecutter/check_cookiecutter.sh
```diff
@@ -70,7 +70,7 @@ function check_templating() {
     local grep_path
     grep_path=$(mktemp)
     grep \
-        --dereference-recursive --context=5 --color=always \
```
wusatosi, line 73: Drive by change so this script work on macOS.

